PR #51, opened 2025-02-13
Add ruff linter

rexposadas:
@noah-duncan 

Up to you if you want this in main. This uses ruff to lint and format.  See: https://github.com/astral-sh/ruff?tab=readme-ov-file#getting-started

I ran `ruff check --fix` to fix linting errors.

I chose ruff because it's fast and it worked well in my last project.

There are other fixes ruff detects, but those need to be done manually.  I didn't do that here to keep the PR specific to automated fixes.

### DIFF
--- a/accounts/admin.py
+++ b/accounts/admin.py
@@ -1,3 +1,2 @@
-from django.contrib import admin
 
 # Register your models here.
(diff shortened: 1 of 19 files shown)


noah-duncan:
Thanks! Always down for some linting

noah-duncan (approved):

PR #56, opened 2025-02-14
handeling edge case with correct error mssg

vishwajeet-hash:
After multiple testing i found the error to be recreated for some scenarios ....mainly when a meeting link with the same session and host in which the bot has been removed ...
If the meeting with same link is started again then again it will work.
In future this case also can be added where rejoining is also possible.

### DIFF
--- a/bots/bot_controller/bot_controller.py
+++ b/bots/bot_controller/bot_controller.py
@@ -487,17 +487,19 @@ class BotController:
             if self.closed_caption_manager:
                 print("Flushing captions...")
                 self.closed_caption_manager.flush_captions()
-
-            if self.bot_in_db.state == BotStates.LEAVING:
-                BotEventManager.create_event(
-                    bot=self.bot_in_db,
-                    event_type=BotEventTypes.BOT_LEFT_MEETING
-                )
+            if self.bot_in_db.state != BotStates.FATAL_ERROR:
+                if self.bot_in_db.state == BotStates.LEAVING:
+                    BotEventManager.create_event(
+                        bot=self.bot_in_db,
+                        event_type=BotEventTypes.BOT_LEFT_MEETING
+                    )
+                else:
+                    BotEventManager.create_event(
+                        bot=self.bot_in_db,
+                        event_type=BotEventTypes.MEETING_ENDED
+                    )
             else:
-                BotEventManager.create_event(
-                    bot=self.bot_in_db,
-                    event_type=BotEventTypes.MEETING_ENDED
-                )
+                print("Bot is in FATAL_ERROR state. Bot cannot rejoin once removed. End the meet for all and start meet again with same link")
             self.cleanup()
             return
 


noah-duncan (changes requested):
Hey, this needed a few changes to make the tests pass. Did you ever run the tests? You can do so locally by running `DJANGO_SETTINGS_MODULE=attendee.settings.test python manage.py test --keepdb --verbosity=2 bots.tests.test_bot_join_meeting.TestBotJoinMeeting.test_bot_in_fatal_error_does_not_create_meeting_ended_event`


I needed to change the test case to this
```
    @patch('deepgram.DeepgramClient')
    @patch('bots.models.BotEventManager.create_event')
    @patch('bots.zoom_bot_adapter.zoom_bot_adapter.zoom', new_callable=create_mock_zoom_sdk)
    @patch('bots.zoom_bot_adapter.zoom_bot_adapter.jwt')
    def test_bot_in_fatal_error_does_not_create_meeting_ended_event(self, mock_create_event, MockDeepgramClient, mock_zoom_sdk_adapter,mock_jwt):
         # Setup mock objects and bot state
        bot = Bot.objects.create(state=BotStates.FATAL_ERROR, project=self.project)
        controller = BotController(bot.id)
        
        # Set up the necessary mocks for Deepgram, Zoom SDK, etc.
        MockDeepgramClient.return_value = create_mock_deepgram()
        
        # Create mock Zoom SDK instance
        mock_zoom_sdk_instance = create_mock_zoom_sdk()
        mock_zoom_sdk_adapter.return_value = mock_zoom_sdk_instance
        # Mock JWT token generation
        mock_jwt.encode.return_value = "fake_jwt_token"

        # Capture logs
        with self.assertLogs(level='INFO') as log:
        # Simulate meeting ended message
            message = {'message': BotAdapter.Messages.MEETING_ENDED}
            controller.bot_in_db = bot
            controller.individual_audio_input_manager = MagicMock()
            controller.closed_caption_manager = MagicMock()
            controller.cleanup = MagicMock()

            # Call the method that handles the message (take_action_based_on_message_from_adapter)
            controller.take_action_based_on_message_from_adapter(message)

            # Assert no event was created for MEETING_ENDED
            mock_create_event.assert_not_called()

            # Check if the log contains the expected message
            self.assertIn('INFO:root:Bot is in FATAL_ERROR state. Bot cannot rejoin once removed. End the meet for all and start meet again with same link', log.output)

            # Verify that cleanup was called
            controller.cleanup.assert_called_once()

            # Ensure the bot's state remains as FATAL_ERROR
            bot.refresh_from_db()
            self.assertEqual(bot.state, BotStates.FATAL_ERROR)

        # Verify expected SDK calls to ensure no meeting creation or interaction was attempted
        mock_zoom_sdk_instance.InitSDK.assert_not_called()
        mock_zoom_sdk_instance.CreateMeetingService.assert_not_called()
        mock_zoom_sdk_instance.CreateAuthService.assert_not_called()
```

And instead of printing, I had to use the logger, to pass the assert in the test
`logging.info("Bot is in FATAL_ERROR state. Bot cannot rejoin once removed. End the meet for all and start meet again with same link")`

bots/bot_controller/bot_controller.py
```diff
@@ -552,10 +553,14 @@ def take_action_based_on_message_from_adapter(self, message):
         if message.get("message") == BotAdapter.Messages.MEETING_ENDED:
             print("Received message that meeting ended")
             self.flush_utterances()
```
noah-duncan, line 567: Were you able to reproduce getting the thing to error out? I couldn't. Anyway, maybe you can have a more generic print message, because this can happen for many different reasons. IE "Message that meeting ended was ignored because bot was already in the FATAL_ERROR state"

bots/bot_controller/bot_controller.py
```diff
@@ -552,10 +553,14 @@ def take_action_based_on_message_from_adapter(self, message):
...
-            if self.bot_in_db.state == BotStates.LEAVING:
-                BotEventManager.create_event(bot=self.bot_in_db, event_type=BotEventTypes.BOT_LEFT_MEETING)
+            if self.bot_in_db.state != BotStates.FATAL_ERROR:
```
noah-duncan: If we can avoid introducing another level of nesting, that'd be good. Can you just have it do
```
if self.bot_in_db.state == BotStates.FATAL_ERROR:
     <log message>
     return
```

Before we get to the other stuff?

Issue #52, opened 2025-02-13
Use PCM for transcription of per-participant audio. Stop converting PCM to MP3


noah-duncan:
The current process for transcribing per participant audio works like this:
1. The `ZoomBotAdapter` sends a chunk of per participant audio to the `IndividualAudioInputManager` by calling the `add_chunk` method. The chunks are in raw PCM format.
2. The `IndividualAudioInputManager` accumulates the chunks and when the participant has stopped speaking for a few seconds, it combines the chunks into an audio clip. The audio clip is sent to the `BotController` by calling the `save_individual_audio_utterance` method. 
3. The `BotController` saves the audio clip as an utterance in the Postgres DB and calls the `process_utterance` background job to transcribe the utterance.
4. The `process_utterance` background job converts the raw PCM to mp3, saves the mp3 in the database, and then calls calls the deepgram API to transcribe the mp3.

@andreapiso pointed out that avoiding the mp3 conversion step altogether reduces the latency of the transcription pipeline. The Deepgram API can also accept raw pcm audio, so we can feed it directly into there.

In this issue, we want to simplify the transcription work done in `bots/tasks/process_utterance_task.py` to stop converting the pcm to mp3 and just send the pcm straight to deepgram. The downside of keeping the raw PCM around is that it takes up alot of space. The only reason to have it around is for debugging purposes. I think we can clear it from the database after transcription. If we encounter an issue where keeping it around is useful, we can easily change the code to keep it for a few days, and then clear it out.

This should be a pretty simple change, just alter a few lines in `bots/tasks/process_utterance_task.py`

noah-duncan:
Closed in https://github.com/attendee-labs/attendee/pull/77

PR #68, opened 2025-02-16
#52 Stop converting PCM to MP3 for transcription

salixnamu:
Instead of having an intermediary step to convert PCM to MP3 before sending over the audio to Deepgram, sends the raw PCM to Deepgram for transcription. 

One thing to do as follow-up is to delete or to have a scheme for deleting the audio after saving the clip


bots/tasks/process_utterance_task.py
```diff
@@ -34,8 +24,7 @@ def process_utterance(self, utterance_id):
     recording = utterance.recording
     RecordingManager.set_recording_transcription_in_progress(recording)
 
-    # if utterance file format is pcm, convert to mp3
-    convert_utterance_audio_blob_to_mp3(utterance)
+    utterance.refresh_from_db() # because of the .tobytes() issue
```
salixnamu: I'm not exactly sure of the .tobytes() issue as presented here but since this is executed async it's prob best to ensure we're using latest state from DB

bots/tasks/process_utterance_task.py
```diff
@@ -47,6 +36,8 @@ def process_utterance(self, utterance_id):
             smart_format=True,
             language=recording.bot.deepgram_language(),
             detect_language=recording.bot.deepgram_detect_language(),
+            encoding="linear16",  # for 16-bit PCM
+            sample_rate=32000, # same sample rate as used in IndividualAudioInputManager
```
salixnamu, line 37: This would be best as a configuration somewhere. I see it set as a class variable for IndividualAudioInputManager so I used the same sampling rate there (changing this number got me differing levels of accuracy with the transcription)

salixnamu:
@noah-duncan I'm unable to assign you to review on GitHub

bots/tasks/process_utterance_task.py
noah-duncan: Yea the comment was super vague. The issue was when calling `utterance.audio_blob`, the type of the object returned when you are loading the column straight from the DB was different from the type of object you get when you previously wrote to that column in-memory by calling `utterance.audio_blob = x` (as happens when you are saving the mp3 data). Adding the `refresh_from_db` ensured that we are always loading straight from the DB. So the return type is always the same. I forget what the two different return types were.

But because we no longer save the mp3 data, we don't need to call refresh_from_db anymore.

bots/tasks/process_utterance_task.py
noah-duncan: Yea, agreed. The most future proof thing to do would probably be to add it as an additional column on the utterance table since it is tied to the audio clip in question. 32Khz is the correct value, but you can verify that it's correct by playing the pcm file at that sample rate and making sure it sounds right. This is a simple program for playing sound files https://www.audacityteam.org/download/

noah-duncan (changes requested):
Hey, looks good just remove that refresh_from_db() we don't need it. And if you want to build more experience with django you can add that column for sample rate. It's not necessary though.

bots/tasks/process_utterance_task.py
salixnamu: I'll add the column and do the migration in a separate PR

bots/tasks/process_utterance_task.py
```diff
@@ -1,20 +1,6 @@
 from celery import shared_task
 from django.db import DatabaseError
 
-from bots.models import Credentials, RecordingManager, Utterance
```
noah-duncan, line 4: Removing this line seems to have broken things. That's why CI is failing. Now that you're a contributor you don't need to fork the repo to make PRs. So in the future, if you make a non-forked PR the CI tests will run automatically.

You can run the CI manually with `DJANGO_SETTINGS_MODULE=attendee.settings.test python manage.py test --keepdb bots.tests`. It doesn't take long to run.

noah-duncan (changes requested):

salixnamu:
Closing this now that I can create non-forked branches

PR #64, opened 2025-02-15
Ruff linter and formatter

rexposadas:
TL;DR: This PR introduces Ruff. Open for discussion.

What This PR Does
	1.	Lints and formats the codebase.
	2.	Adds pyproject.toml to configure Ruff settings.

pyproject.toml includes:
	•	Exclusions for certain directories (e.g., migrations).
	•	Disabled specific rules as needed.

Notes:
	1.	Adds .pre-commit-config.yaml to enforce linting/formatting before commits.
	2.	git commit will fail if files are not linted/formatted.
	3.	To bypass the linter, use git commit -n.

A separate PR will introduce Ruff checks in CI, ensuring PRs cannot pass without compliance.

PR #77, opened 2025-02-23
Stop PCM conversion and add sample_rate column to utterances

salixnamu:
Instead of having an intermediary step to convert PCM to MP3 before sending over the audio to Deepgram, sends the raw PCM to Deepgram for transcription. Also sets a new variable, `sample_rate` which is needed for Deepgram to parse PCM audio accurately.

This `sample_rate` is saved for each utterance in the DB, with default value of 32K.


bots/tasks/process_utterance_task.py
```diff
@@ -64,6 +52,7 @@ def process_utterance(self, utterance_id):
 
         response = deepgram.listen.rest.v("1").transcribe_file(payload, options)
         utterance.transcription = json.loads(response.results.channels[0].alternatives[0].to_json())
+        utterance.audio_blob = b'' # set the binary field to empty byte string
```
salixnamu: Don't save PCM utterances since they're so large. By default the column does not accept null values so went with empty byte string
Screenshot 2025-02-24 at 9 40 31 PM

bots/models.py
```diff
@@ -713,6 +713,7 @@ class AudioFormat(models.IntegerChoices):
     duration_ms = models.IntegerField()
     transcription = models.JSONField(null=True, default=None)
     source_uuid = models.CharField(max_length=255, null=True, unique=True)
+    sample_rate = models.IntegerField(default=32000)
```
noah-duncan: Forgot to mention this, but the Utterance class is also used in Google Meet bots. For Google Meet we are not associating any raw audio with the utterance, because we got the transcribed utterance from Google Meet's closed captions. So this column should be nullable, and in this line where the Google Meet based Utterance gets created, you should set the sample_rate to null.

```
        utterance, _ = Utterance.objects.update_or_create(
            recording=recording_in_progress,
            source_uuid=source_uuid,
            defaults={
                "source": Utterance.Sources.CLOSED_CAPTION_FROM_PLATFORM,
                "participant": participant,
                "transcription": {"transcript": message["text"]},
                "timestamp_ms": message["timestamp_ms"],
                "duration_ms": message["duration_ms"],
            },
        )
 ```

**Review by noah-duncan — CHANGES_REQUESTED** [review]

File: bots/bot_controller/bot_controller.py [hunk 3]
```diff
@@ -445,13 +445,7 @@ def save_closed_caption_utterance(self, message):
...
             recording=recording_in_progress,
             source_uuid=source_uuid,
-            defaults={
-                "source": Utterance.Sources.CLOSED_CAPTION_FROM_PLATFORM,
-                "participant": participant,
-                "transcription": {"transcript": message["text"]},
-                "timestamp_ms": message["timestamp_ms"],
-                "duration_ms": message["duration_ms"],
-            },
+            defaults={"source": Utterance.Sources.CLOSED_CAPTION_FROM_PLATFORM, "participant": participant, "transcription": {"transcript": message["text"]}, "timestamp_ms": message["timestamp_ms"], "duration_ms": message["duration_ms"], "sample_rate": None},
```
salixnamu: Linter reduced this to one line, I'm not sure if want to change the rules here

salixnamu: Hey @rexposadas, I saw that you updated the linter to no longer break lines in #64. Is this what we want for large multiline object definitions like the above? Just verifying this is intentional, because it does affect readability

noah-duncan: Yea it's annoying. Looked into it and there didn't seem to be a way to tell ruff to just never collapse lines. It's a super duper popular linter so its surprising. Maybe I'm missing something.

It does seem that if you put a trailing comma at the end of your dict, IE
Instead of
`{"source": Utterance.Sources.CLOSED_CAPTION_FROM_PLATFORM, "participant": participant, "transcription": {"transcript": message["text"]}, "timestamp_ms": message["timestamp_ms"], "duration_ms": message["duration_ms"], "sample_rate": None}`
do

`{"source": Utterance.Sources.CLOSED_CAPTION_FROM_PLATFORM, "participant": participant, "transcription": {"transcript": message["text"]}, "timestamp_ms": message["timestamp_ms"], "duration_ms": message["duration_ms"], "sample_rate": None,}`

Then it keeps the line breaks.

noah-duncan (approved):

rexposadas:
@salixnamu  sorry I didn't address your comment in a timely manner. I missed the notification when you tagged my name.  That's strange. I usually get my notifications. 

Thanks @noah-duncan for addressing it.

PR #95, opened 2025-02-27
Load video duration at start in bot detail page

noah-duncan:
The videos now stream immediately, but we can't see their duration. This PR uses videojs to improve the player experience in the bot detail page. You'll see the video duration immediately and be able to move through the timeline however you want.


noah-duncan:
Closed in favor of the approach taken by https://github.com/noah-duncan/attendee/pull/96

PR #97, opened 2025-02-28
Add setup_test_db command to Django management

salixnamu:
While setting up local environment for testing, found that
`DJANGO_SETTINGS_MODULE=attendee.settings.test python manage.py test --keepdb bots.tests`
would fail because the postgres database container does not come with the test database and test user.

Added a `setup_test_db` command to be run before running the above if you don't have the necessary db state. Run
`DJANGO_SETTINGS_MODULE=attendee.settings.test python manage.py setup_test_db`
to create the db and user and grant privileges.

If you don't have these setup, it'll look like below:
Screenshot 2025-02-28 at 3 21 47 PM
And if you do have the setup,
Screenshot 2025-02-28 at 3 22 11 PM

I welcome any feedback on a location in particular where this documentation should be added

### DIFF
--- a/bots/management/commands/setup_test_db.py
+++ b/bots/management/commands/setup_test_db.py
@@ -1,0 +1,99 @@
+import os
+
+import django
+
+# Set the default Django settings module
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "attendee.settings.test")
+
+# Initialize Django (this is required before accessing settings)
+django.setup()
+
+import psycopg2
+from django.conf import settings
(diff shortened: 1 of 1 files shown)


rexposadas:
As far as documentation location, somewhere near the section dealing with local setup might be a good: https://github.com/noah-duncan/attendee?tab=readme-ov-file#running-in-development-mode

Just a thought.

noah-duncan (approved):
Thanks @salixnamu looks good! And as @rexposadas adding directions to the local setup section would be good.

PR #106, opened 2025-03-06
updated deprecated call to utctime

rexposadas:
`datetime.utcnow()` has been deprecated.

See: https://docs.python.org/3/library/datetime.html#datetime.datetime.now

This PR updates the codebase accordingly.

### DIFF
--- a/bots/bot_controller/closed_caption_manager.py
+++ b/bots/bot_controller/closed_caption_manager.py
@@ -5,24 +5,25 @@ from typing import Dict, Optional
 class CaptionEntry:
     def __init__(self, caption_data: dict):
         self.caption_data = caption_data
-        self.created_at = datetime.utcnow()
+        self.created_at = datetime.now(datetime.UTC)
         self.modified_at = self.created_at
         self.last_upsert_to_db_at: Optional[datetime] = None
 
     def update(self, caption_data: dict):
         self.caption_data = caption_data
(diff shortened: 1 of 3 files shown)


rexposadas:
@noah-duncan  -- still working on this one. not sure why removing a deprecated call broke things.

noah-duncan:
If you look in the test logs you can see `AttributeError: type object 'datetime.datetime' has no attribute 'UTC'`

rexposadas:
Got it. The logs indicate that we’re using Python 3.10, but our .python-version file and ci.yml configuration specify Python 3.12.

Example from the logs:

```
File "/usr/lib/python3.10/threading.py", line 1378, in run  
    self.function(*self.args, **self.kwargs)  
```

Versus our CI configuration:
[ci.yml line 83](https://github.com/noah-duncan/attendee/blob/main/.github/workflows/ci.yml#L83)

Which version should we standardize on?

noah-duncan:
Okay. Our dockerfile's base image is ubuntu:22.04, which comes with python 3.10 pre-installed. 

For me, the path of least resistance would just be to go with what the dockerfile is doing, so python 3.10. I haven't put a great deal of thought into this, so lmk if you see reasons to do 3.12.

rexposadas:
Low-priority. Not needed now.

PR #116, opened 2025-03-13
Add webhook subscriptions

salixnamu:
This pull request adds functionality for webhook subscriptions to the `attendee` project. The following key changes were made:

1. **Models**:
   - Added `WebhookSecret`, `WebhookSubscription`, and `WebhookDeliveryAttempt` models to store webhook configurations and delivery attempts.
   - Introduced `WebhookTriggerTypes` and `WebhookDeliveryAttemptStatus` to manage webhook event types and statuses.

2. **Views**:
   - Created `ProjectWebhooksView` to display existing webhooks for a project.
   - Implemented `CreateWebhookView` to handle the creation of new webhook subscriptions.
   - Added `DeleteWebhookView` to delete existing webhook subscriptions.
   - NOTE: All views are meant for authenticated users, like the Get/Create/Delete ApiKey flows

3. **Tasks**:
   - Introduced a Celery task `deliver_webhook` to handle the delivery of webhooks with retry logic and logging.


bots/bots_api_views.py
```diff
@@ -667,3 +671,48 @@ def get(self, request, object_id):
...
+        if WebhookSubscription.objects.filter(url=serializer.validated_data["url"]).exists():
+            return Response({"error": "URL already subscribed"}, status=status.HTTP_400_BAD_REQUEST)
+
+        # Get the project's secret for the webhook subscription. If new project, create a new one
+        webhook_secret = WebhookSecret.objects.filter(project=request.auth.project).first()
+        if not webhook_secret:
+            webhook_secret = WebhookSecret.objects.create(
+                project=request.auth.project
+            )
+            webhook_secret.set_secret(secrets.token_bytes(32))
```
noah-duncan: I assume anytime a webhook secret is created, you'll want to initialize the secret value. So you can put this in the model itself by overriding the save method. See how we set the object_id for various classes.

```
    def save(self, *args, **kwargs):
        if not self.object_id:
            # Generate a random 16-character string
            random_string = "".join(random.choices(string.ascii_letters + string.digits, k=16))
            self.object_id = f"{self.OBJECT_ID_PREFIX}{random_string}"
```

bots/migrations/0017_webhooksecret_webhooksubscription_and_more.py
```diff
@@ -0,0 +1,61 @@
+# Generated by Django 5.1.2 on 2025-03-15 01:12
```
noah-duncan: I had this long-standing PR to add credits for organization and it was just merged into master. It was called 0017, so you need to reverse this migration locally, then delete the file, then recreate the file to fix the conflict. sorry about that!

noah-duncan (changes requested):

bots/bots_api_views.py
```diff
@@ -667,3 +670,49 @@ def get(self, request, object_id):
 
         except Bot.DoesNotExist:
             return Response({"error": "Bot not found"}, status=status.HTTP_404_NOT_FOUND)
+
+
+class WebhookSubscriptionView(APIView):
+    authentication_classes = [ApiKeyAuthentication]
+
+    @extend_schema(
```
noah-duncan: Sorry, should've spotted this earlier. To follow the precedent in the rest of the codebase for how we create objects that are only createable in the GUI, a webhook subscription like this should not be created via an API key. It'll be created via cookie in the graphical UI. See how we create things like credentials and api keys. It's all in `bots/projects_views.py`. The URL for creating webhook subscriptions should be in `bots/projects_urls.py`.

bots/bots_api_urls.py
```diff
@@ -3,6 +3,11 @@
 from . import bots_api_views
 
 urlpatterns = [
+    path(
+        "webhook_subscription",
+        bots_api_views.WebhookSubscriptionView.as_view(),
```
noah-duncan: This should be in `bots/projects_urls.py`, it's not a true api endpoint since you create it in the UI. See the comment above for more detail.

bots/webhook_utils.py
```diff
@@ -0,0 +1,59 @@
...
+logger = logging.getLogger(__name__)
+
+
+def trigger_webhook(webhook_event_type, bot, payload):
+    """
+    Trigger a webhook for a given event.
+    """
+    from bots.models import WebhookDeliveryAttempt, WebhookSubscription
+
+    subscriptions = WebhookSubscription.objects.filter(events__contains=[webhook_event_type], is_active=True)
```
noah-duncan: You're querying across all the webhook subscriptions in the database. It should only be for the bot's project. We should probably have a test to catch this, because it's a big deal to leak data. Maybe there's some kind of abstraction in django we can use to make it harder to write these kinds of queries.

bots/models.py
```diff
@@ -1040,3 +1055,85 @@ def url(self):
...
+
+
+class WebhookSubscription(models.Model):
+    def default_events():
+        return [WebhookTriggerTypes.BOT_STATE_CHANGE]
+
+    project = models.ForeignKey(Project, on_delete=models.CASCADE, related_name="webhook_subscriptions")
+    url = models.URLField()
+    events = models.JSONField(default=default_events)
+    secret = models.ForeignKey(WebhookSecret, on_delete=models.PROTECT, related_name="webhook_subscriptions")
```
noah-duncan: Do we need the foreign key for subscription <--> secret or can we just say that we'll sign all webhook payloads with ALL the active secrets that are currently associated to the subscription's project? That allows for secret rotation. Under this schema, if we rotate or delete a secret, then this foreign key will have to be updated which I think adds unneeded complication.

bots/models.py
```diff
@@ -1040,3 +1055,85 @@ def url(self):
...
+    is_active = models.BooleanField(default=True)
+    reputation = models.IntegerField(default=0)
```
noah-duncan: This doesn't seem to be used anywhere, so maybe we add when we have a use for it?

bots/tasks/deliver_webhook_task.py
```diff
@@ -0,0 +1,90 @@
...
+    except WebhookDeliveryAttempt.DoesNotExist:
+        logger.error(f"Webhook delivery attempt {delivery_id} not found")
+        return
+
+    subscription = delivery.webhook_subscription
+
+    # If the subscription is no longer active, mark as failed and return
+    if not subscription.is_active:
+        delivery.status = "failed"
+        delivery.error_message = "Webhook subscription is no longer active"
```
noah-duncan: Did you forget to add the `error_message` column in the migration? Maybe the `response_bodies_list` column can be generalized to include this kind of thing.

bots/tasks/deliver_webhook_task.py
```diff
@@ -0,0 +1,90 @@
...
+        delivery = WebhookDeliveryAttempt.objects.get(id=delivery_id)
+    except WebhookDeliveryAttempt.DoesNotExist:
+        logger.error(f"Webhook delivery attempt {delivery_id} not found")
+        return
+
+    subscription = delivery.webhook_subscription
+
+    # If the subscription is no longer active, mark as failed and return
+    if not subscription.is_active:
+        delivery.status = "failed"
```
noah-duncan: delivery.status is an integer, not a string. You should add a test that would catch this

bots/tasks/deliver_webhook_task.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # Send the webhook
+    try:
+        response = requests.post(
+            subscription.url,
+            json=webhook_data,
+            headers={
+                "Content-Type": "application/json",
+                "User-Agent": "Attendee-Webhook/1.0",
+                "X-Webhook-Signature": signature,
```
noah-duncan, line 67: For the case of webhook secret rotation when two secrets are active at the same time, we'd add both of them, separated by spaces, like [svix does](https://www.svix.com/blog/zero-downtime-secret-rotation-webhooks/#zero-downtime-secret-rotation). This is fine for now, let's not worry about rotation aside from making sure the DB schema will let us add it in the future.

bots/tasks/deliver_webhook_task.py
```diff
@@ -0,0 +1,90 @@
...
+            },
+            timeout=10,  # 10-second timeout
+        )
+
+        # Update the delivery attempt with the response
+        delivery.response_status_code = response.status_code
+
+        # Limit response body storage to prevent DB issues with large responses
```
noah-duncan, line 75: good call

bots/tasks/deliver_webhook_task.py
```diff
@@ -0,0 +1,90 @@
...
+            delivery.save()
+            return
+
+        # If we got here, the delivery failed with a non-2xx status code
+        delivery.status = WebhookDeliveryAttemptStatus.FAILURE
+
+    except requests.RequestException:
+        # Handle network errors, timeouts, etc.
+        delivery.status = WebhookDeliveryAttemptStatus.FAILURE
+
```
noah-duncan, line 99: can you store the exception in the error message field

noah-duncan: Maybe generalize response_body_list to also store this kind of data in addition to http responses. If we don't get an http response we'll get an exception of some kind. This is a good place to store it.

noah-duncan (changes requested):
Hey looked into it in more depth and added more feedback. Nothing that'll take that long to fix.

bots/tasks/deliver_webhook_task.py
salixnamu: I changed the status to the int for code and added the message into the `response_bodies_list` but now I'm wondering if we should do this for inactive subscriptions. No delivery attempt is made, and I'm not sure that should mean the attempt _failed_, and the length of `response_bodies_list` would differ from the number in `attempt_count` if there were some webhook deliveries that passed over an inactive subscription

bots/tasks/deliver_webhook_task.py
salixnamu: I'll keep a note that secret rotation should be a quick followup to the initial webhook feature

bots/bots_api_views.py
salixnamu: UI has been added to this PR, follows very similar flow as for creating and deleting the API keys.

bots/tasks/deliver_webhook_task.py
salixnamu: response_body_list is just a list so I suppose we could append the string of the exception message into the list

bots/templates/projects/sidebar.html
```diff
@@ -1,116 +1,144 @@
...
-                    <hr>
-                    <form method="post" action="{% url 'account_logout' %}" class="m-0">
-                        {% csrf_token %}
-                        <button type="submit" class="nav-link border-0 bg-transparent w-100 text-start">
-                            <i class="bi bi-box-arrow-right"></i> Logout
-                        </button>
-                    </form>
-                </div>
-            </div>
+  <div class="row">
```
salixnamu: Ugh I didn't mean to do this. I have some prettier settings enabled on my local vscode. Running ruff linter doesn't seem to affect templates. I can go back later and see about disabling my prettier settings and only making the change I need

Or better yet set some sort of configuration for the templates as well

salixnamu:
Tests are involving a little more updates than expected, these will be resolved tomorrow

bots/models.py
salixnamu: I removed the foreign key. Although we do not yet support secret rotation, for now we have secrets that are only associated with the project, not any webhook subscription

noah-duncan (approved):
Thanks, this was a major effort and is going to make one of our users very happy :) Great work!

PR #142, opened 2025-03-26
Improve gmeet join logic

noah-duncan:


noah-duncan:
Not sure why that last test won't pass in the cloud. Need to move on to other things, so pushing up a truncated pr instead.

PR #147, opened 2025-03-26
Joining error failing fix

### LINKED ISSUE #108: The microsoft teams bot joining takes place multiple times.
Even when a bot is already in lobby another bot also joins the meet...
I have seen once 3 bots waiting in lobby...

vishwajeet-hash:
Fix #108 
For turning on the captions in ms teams flow currently it is necessary that the mic of the bot entering is turned on as it joins the meet ...
This was one of the reason for failing at times while joining process..
That case is being handled here..

### DIFF
--- a/bots/teams_bot_adapter/teams_ui_methods.py
+++ b/bots/teams_bot_adapter/teams_ui_methods.py
@@ -117,7 +117,7 @@ class TeamsUIMethods:
 
         self.fill_out_name_input()
 
-         # Check if the mic is off and turn it on if needed
+        # Check if the mic is off and turn it on if needed
         logger.info("Checking if the microphone is off...")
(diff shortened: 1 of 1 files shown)


noah-duncan (changes requested):
Thanks for the PR. Currently, it does not work when running the bot in a docker container. In a docker container there is no microphone to be enabled. See the screenshot:

image
```
Message: element click intercepted: Element <div '
                            'class="ui-box cx cy cz da db gk gl gm gn go dh di '
                            'dj dk dl dm dn do dp dq dr ds dt cn du dv dw dx '
                            'dy dz ea eb gp gq gr gs fo fp fq fr fs ft fu fv '
                            'ec fw ee fx ui-checkbox__indicator"></div> is not '
                            'clickable at point (550, 551).
                            ```
                            
You need to change the code so that it only attempts to click if it the toggle is enabled. When in the docker container it is disabled.

PR #166, opened 2025-04-11
feat: Add S3 recording storage path configuration and update file uploader

DaviReisVieira:
Introduce a configuration for the S3 recording storage path and modify the file uploader to support this path when uploading files.

- [x] Added new optional variable `AWS_RECORDING_STORAGE_PATH`


noah-duncan:
Hi @DaviReisVieira can you talk about why you need this?

DaviReisVieira:
Hi, @noah-duncan ! Thanks for the question!

The motivation behind introducing`AWS_RECORDING_STORAGE_PATH` is to support more flexible and organized file storage within the S3 bucket. Teams prefer to group recordings under a specific folder/prefix (e.g., per environment, customer, or date) instead of storing all recordings at the root of the bucket.

This optional config allows us to define a path like recordings/meetings/ or prod/recordings/, which helps:
- Keep S3 buckets organized and scalable
- Avoid potential filename collisions
- Make it easier to apply lifecycle rules or access policies by prefix
- Enable multi-tenant or environment-specific storage patterns

It’s backward-compatible — if the path isn’t defined, it behaves exactly as before.

Let me know if you’d like this refactored or split in any way!

noah-duncan:
Hi @DaviReisVieira Sorry for taking a long time to reply. I don't see a scenario where you need to put the recordings inside a folder within an existing bucket, instead of just creating a new bucket for only recordings. I'm going to close this PR for now, feel free to DM me in the slack if you want to talk about it more.

PR #191, opened 2025-04-22
nove-3 supports multilingual; add env to replace the bots docker img

steinathan:


bots/tasks/process_utterance_task.py
```diff
@@ -35,11 +35,7 @@ def process_utterance(self, utterance_id):
             "buffer": utterance.audio_blob.tobytes(),
         }
 
-        # nova-3 does not have multilingual support yet, so we need to use nova-2 if we're transcribing with a non-default language
-        if (recording.bot.deepgram_language() != "en" and recording.bot.deepgram_language()) or recording.bot.deepgram_detect_language():
-            deepgram_model = "nova-2"
-        else:
-            deepgram_model = "nova-3"
+        deepgram_model = "nova-3-general"
```
steinathan, line 38: or maybe `nova-3-meeting`?

https://developers.deepgram.com/docs/model

noah-duncan: Yea sure, we can try that. So you tested nova-3 with other languages and it works?

noah-duncan: Hmm, it looks like if you pass a specific language, then you need to use nova-2. Unclear if deepgram wants you to pass language=multi and use nova-3 for everything or fallback to nova-2 when it's a known non-english language. For now, this logic is the most fullproof: https://github.com/noah-duncan/attendee/pull/190/files#diff-3594204af70953243d0a359357257ec0b5df9e4e5b48bd8e65199ac8a4dba353R44

steinathan: yes you're right, depegram wants us to pass `model=nova-3` and use `language=multi` 

https://developers.deepgram.com/docs/multilingual-code-switching

PR #232, opened 2025-05-14
handle join denied case for Teams

dourgulf:
Change click_captions_button, use short timeout and try, check denied message then every timeout.


noah-duncan (approved):
Thanks, this looks good. The only issue is that you should wait for a time equal to `automatic_leave_configuration.waiting_room_timeout_seconds`. The previous code was doing that, but the new code isn't. I'll merge this in and fix it in another PR.

noah-duncan:
Actually, I wanted to fix the linting error before merging, so made a new branch. Will close this PR. Thanks for the contribution!

PR #235, opened 2025-05-16
Add websocket support for bot audio/video control

igui:
- Added `BotWebsocketClient` to manage websocket connections for sending and receiving messages.
- Updated `BotController` to handle websocket interactions, including sending audio chunks and event notifications.
- Added some new event types and subtypes for websocket communication.
- Added some options to  `PipelineConfiguration` to include websocket streaming options.
- Modified command line arguments for launching the bot to support websocket settings.

TODO
- I still need to test it with a Teams meeting and a Google Meet meeting
- I would need to test it in a smaller real instance, as processing many websocket events can cause the bot to run behind in the queue.

### Testing
For working, this code requires a websocket server implementation that streams audio from/back to the bot; for example, an AI bot. I can add one as an example in another file if you like.

To test it you might run the following command:

```shell
curl http://localhost:8000/api/v1/bots \
  --request POST \
  --header 'Authorization: Token SOME_TOKEN' \
  --header 'Content-Type: application/json' \
  --data '{
    "meeting_url": "SOME_URL",
    "bot_name": "Meeting Bot",
    "audio_websocket_url": "ws://172.20.0.1:39305"
  }'
```

If we want to have a sample websocket server implementation, check out this repo https://github.com/igui/attendee-deepgram-websockt-example which has a sample implementation of the Websocket server that is compatible with Attendee.


bots/bot_controller/bot_controller.py
```diff
@@ -737,6 +768,27 @@ def save_debug_recording(self):
...
+        if not self.websocket_client:
+            return
+        self.websocket_client.send_async(
+            {
+                "event_type": event_type.name,
+                "event_sub_type": event_sub_type.name if event_sub_type else None,
+                "event_metadata": event_metadata,
+                "event_data": event_data,
+            }
+        )
```
igui: This function has to be super fast (non-blocking) as it is possibly called from a GLib main loop. No JSON serialization or any IO calls should happen here.

Also, buffering might be necessary, for example, sending audio in larger batches of 512 bytes.

bots/bot_controller/bot_controller.py
```diff
@@ -943,6 +1059,9 @@ def take_action_based_on_message_from_adapter(self, message):
                 bot=self.bot_in_db,
                 event_type=BotEventTypes.BOT_RECORDING_PERMISSION_GRANTED,
             )
+            self.send_message_to_websocket(
+                event_type=BotEventTypes.BOT_RECORDING_PERMISSION_GRANTED,
+            )
```
igui: Not entirely necessary but nice to have.
When implementing a conversational bot, they send a greeting when the conversation starts, like "Hello! how can I help you". 

The audio will only be heard if the bot has mic permissions.

igui: One can implement it by having a webhook listening  *and* a websocket (only for audio), but it implies having two things set up, which I really dislike.

bots/models.py
```diff
@@ -255,6 +255,13 @@ def rtmp_destination_url(self):
 
         return f"{destination_url}/{stream_key}"
 
+    def websocket_url(self):
+        """Websocket URL is used to send video frames to the bot"""
+        websocket_settings = self.settings.get("websocket_settings", {})
+        if websocket_settings is None:
+            return None
+        return websocket_settings.get("url", None)
```
noah-duncan: it would be more flexible to have a separate websocket url for each type of data ie `audio_input_url` `audio_output_url`, `state_changes_url`, `transcription_url`

igui: Ok, then I changed and I left and `audio_url` for the audio media. Web sockets have two streams of data (input and output) so we can use Audio.

bots/bot_controller/bot_controller.py
noah-duncan: Yea, that's fair. In the current state of this PR you have to setup both, but the ground work is being laid to send all webhook events over websockets by using a consistent schema for both audio packets and webhooks.

bots/bot_controller/bot_controller.py
noah-duncan: You need buffering for zoom specifically. If you use zooms raw audio sending method and send too much audio at once, it causes zoom's internal buffer to overflow and the audio will "skip ahead". I fixed this problem in `bots/bot_controller/realtime_audio_output_manager.py` by having another buffer outside of zoom and slowly feeding the audio packets from this buffer to zoom, so that zoom does not get overwhelmed.

steinathan:
@noah-duncan is this repo still valid for these changes? is there an updated code example?

https://github.com/igui/attendee-deepgram-websockt-example/tree/main

noah-duncan:
No, the format for the websocket payloads is different. I'm working on an updated example now.

PR #246, opened 2025-05-27
feat: add support for maximum up time the bot should be allowed to run

steinathan:
There are scenarios where allowing a bot to run indefinitely is not desirable. For example, when providing access to trial accounts, demo environments, or limited user tiers, we may want to enforce time-based constraints to control usage, prevent resource abuse, and better manage system load.

By introducing a configurable `max_uptime_seconds` field, we can explicitly define how long a bot should remain active before automatically terminating. ignoring or setting this value to None by default retains flexibility

----
- Introduced `max_uptime_seconds` field with a default of None to allow unlimited bot uptime
- Updated schema description and default to reflect new behavior


noah-duncan:
Thanks this is a needed feature, it looks good to go from first glance, will review more thoroughly tomorrow.

steinathan:
@noah-duncan  Sure, thanks - take your time


oh yeah, i previously used `float(inf)` but django didnt like that, so i used `None` instead and checked against that

noah-duncan:
Ah ok got it. Just added you as a collaborator to the repo, so you won't have to fork in the future. 

Also, the linting CI check is currently failing. Just run `ruff check --fix; ruff format` to lint your code so the CI passes.

steinathan:
Thanks, fixed

noah-duncan (approved):
Thanks! Looks good

PR #249, opened 2025-05-27
handle case: Zoom host switch Bot between “Put in waiting room" and "Admit" 

dourgulf:
Would you please do some checking about the recording action? There remains some error warning which I am not sure how to fix it, but works.


dourgulf:
maybe still not good enough to handle such case, close it

PR #283, opened 2025-06-23
feat: extend openai transcription with custom endpoint and model names

jeremytregunna:
This PR extends the OpenAI transcription configuration via environment variables, allowing users to set:

- `OPENAI_BASE_URL`
- `OPENAI_MODEL_NAME`

to custom values, such that when transcription is configured with an OpenAI key, and these environment variables are set, the transcription service will call into the custom endpoint, with the custom model name to perform the transcription.

This PR does not change any default behaviour, defaulting to `https://api.openai.com/v1` as the default base url, and `gpt-4o-transcribe` as the default model name.

### DIFF
--- a/bots/models.py
+++ b/bots/models.py
@@ -257,7 +257,8 @@ class Bot(models.Model):
         return self.settings.get("transcription_settings", {}).get("openai", {}).get("prompt", None)
 
     def openai_transcription_model(self):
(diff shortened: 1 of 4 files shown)


noah-duncan:
Thanks for the contribution! I needed to make a few tweaks to get the tests, going to close this  and make another PR with those added.

PR #291, opened 2025-06-26
Update recording filename format to include bot ID

docteurZ:
- Modified get_recording_filename() to include both bot object_id and recording object_id
- Provides better traceability and context for recording files in S3/GCS storage


noah-duncan (approved):
Thanks for the idea and the PR!

PR #296, opened 2025-06-28
Feature/bot level webhook subscriptions

docteurZ:
This enables users to create webhook subscriptions specific to individual bots  via the API, while maintaining backward compatibility with existing project-level webhooks managed through the UI. Bot-level webhooks take priority when present.


bots/bots_api_utils.py
```diff
@@ -134,6 +137,7 @@ def create_bot(data: dict, source: BotCreationSource, project: Project) -> tuple
     bot_chat_message = serializer.validated_data["bot_chat_message"]
     metadata = serializer.validated_data["metadata"]
     join_at = serializer.validated_data["join_at"]
+    webhook_subscriptions = serializer.validated_data["webhook_subscriptions"]
```
noah-duncan: Minor thing but we refer to webhook subscriptions as "webhooks" in the UI to simplify the verbiage. So to maintain consistency, I would call them "webhooks" in the API call parameter as well.

Naming the variable `webhook_subscriptions` is good though, because that is what they are called in the backend codebase. So I would change this line to
`webhook_subscriptions = serializer.validated_data["webhooks"]` and change the serializer accordingly.

bots/models.py
```diff
@@ -1724,12 +1724,44 @@ def trigger_type_to_api_code(cls, value):
         }
         return mapping.get(value)
 
+    @classmethod
+    def api_code_to_trigger_type(cls, api_code):
```
noah-duncan: Instead of creating a second dict in `trigger_type_to_api_code`,  I would define one dict:
```
{
            cls.BOT_STATE_CHANGE: "bot.state_change",
            cls.TRANSCRIPT_UPDATE: "transcript.update",
            cls.CHAT_MESSAGES_UPDATE: "chat_messages.update",
            cls.PARTICIPANT_EVENTS_JOIN_LEAVE: "participant_events.join_leave",
}
 ```
 
 And then reference it in both `trigger_type_to_api_code` and `api_code_to_trigger_type`. So when we add a trigger, we don't have to update two dicts.

File: bots/serializers.py [hunk 3]
```diff
@@ -419,6 +419,12 @@ class CreateBotSerializer(serializers.Serializer):
     metadata = MetadataJSONField(help_text="JSON object containing metadata to associate with the bot", required=False, default=None)
     bot_chat_message = BotChatMessageRequestSerializer(help_text="The chat message the bot sends after it joins the meeting", required=False, default=None)
     join_at = serializers.DateTimeField(help_text="The time the bot should join the meeting. ISO 8601 format, e.g. 2025-06-13T12:00:00Z", required=False, default=None)
+    webhook_subscriptions = serializers.ListField(
```
noah-duncan: The pattern we've followed for other complex attributes is to define another serializer class that descends from `serializers.JSONField` just for that attribute. 

See how `TranscriptionSettingsJSONField` is used for example. The benefit of this approach is you can apply JSON schema validation which gets you a bunch of error messages for free.

So to follow that pattern you'd define a class called `WebhooksJSONField` and define the exact schema for valid JSON describing the webhooks.

bots/serializers.py
```diff
@@ -771,6 +777,48 @@ def validate_join_at(self, value):
...
+            return value
+
+        if not isinstance(value, list):
+            raise serializers.ValidationError("webhook_subscriptions must be a list")
+
+        for webhook_data in value:
+            if not isinstance(webhook_data, dict):
+                raise serializers.ValidationError("Each webhook subscription must be a dictionary")
+
+            if "url" not in webhook_data:
```
noah-duncan: If you look at the usage of `@extend_schema_field` in the file you can see we're using [JSON-schema](https://json-schema.org/) to implictly create these kinds of validations, instead of writing them using normal python code.

docs/webhooks.md
```diff
@@ -2,18 +2,92 @@
 
 Webhooks send your server real-time updates when something important happens in Attendee, so that you don't need to poll the API.
 
-Currently, webhooks are only supported for one type of event: when a bot changes its state. This can be used to alert your server when a bot joins a meeting, starts recording or when a recording is available.
```
noah-duncan, line 5: ~~This is annoying, yesterday, I added another type of webhook trigger for participant events. See this PR: https://github.com/attendee-labs/attendee/pull/292~~

~~It looks like when you added your edits to the webhooks documentation, it was using an old version of the webhook docs file that didn't include that. I wonder why github isn't registering a merge conflict.~~

bots/bots_api_utils.py
```diff
@@ -176,8 +180,116 @@ def create_bot(data: dict, source: BotCreationSource, project: Project) -> tuple
...
+    else:
+        # For project-level webhooks, check if URL already exists for project
+        if existing_webhook_query.filter(bot__isnull=True).exists():
+            return "URL already subscribed", None
+
+    # Webhook limit check (only if check_limits is True)
+    if check_limits:
+        if bot:
+            # For bot-level webhooks, check the limit per bot (max 2 per bot)
+            bot_level_webhooks = WebhookSubscription.objects.filter(project=project, bot=bot).count()
```
noah-duncan: I think you could put the bot-level limit check in the serializer. Just inspect the JSON passed to the API.

bots/projects_views.py
```diff
@@ -396,30 +397,26 @@ def post(self, request, object_id):
...
+            # Map specific error messages for project-level webhooks to maintain backward compatibility
+            if "URL already subscribed for this bot" in error:
+                error = "URL already subscribed"
+            return HttpResponse(error, status=400)
 
-        # Get the project's secret for the webhook subscription. If new project, create a new one
-        webhook_secret, created = WebhookSecret.objects.get_or_create(project=project)
+        # Create webhook subscriptions using shared function
+        webhook_data = [{"url": url, "triggers": normalized_triggers}]
+        success, error = create_webhook_subscriptions(webhook_data, project, bot=None, skip_validation=True)
```
noah-duncan: This skip_validation parameter seems a bit clunky. Ideally we would just validate when we create, and not have different logic for project webhook vs bot webhook.

Would it simplify things if instead of `create_webhook_subscriptions`, we had a `create_webhook_subscription` method that is only responsible for creating a single webhook subscription? And then when we need to create multiple for the bot, we just loop over it?

docs/webhooks.md
noah-duncan: Whoops, you weren't using an old version of the webhook docs file. The latest version of the file just had this incorrect line: `Currently, webhooks are only supported for one type of event: when a bot changes its state.`

Thanks for updating it!

docs/webhooks.md
```diff
@@ -2,18 +2,92 @@
...
+Webhooks support multiple event types including bot state changes, transcript updates, and chat message updates. These can be used to alert your server when a bot joins a meeting, starts recording, when a recording is available, or when real-time content is updated.
```
noah-duncan: also mention participant events

docs/webhooks.md
```diff
@@ -2,18 +2,92 @@
...
+- `url` (required): HTTPS URL to receive webhook events
+- `triggers` (required): Array of webhook trigger strings: `["bot.state_change", "transcript.update", "chat_messages.update"]`
+
+### Available Webhook Triggers
+
+| Trigger | Description |
+|---------|-------------|
+| `bot.state_change` | Bot changes state (joins, leaves, starts recording, etc.) |
+| `transcript.update` | Real-time transcript updates during meeting |
+| `chat_messages.update` | Chat message updates in the meeting |
```
noah-duncan, line 53: Also mention participant events

PR #326, opened 2025-07-17
Fix JSON parsing in Teams WebRTC data handlers

wagao29:
Teams WebRTC messages contain binary data followed by JSON data. Binary data can contain bytes that match [ or { characters, causing the JSON parser to fail when it tries to parse from the wrong position.

### Example case
```
O�mL������T}�	��R�	'){�%P�[{"history":[3609,3396,3183,2970,2331,840,2118,1053,414,1905],"type":"dsh"}]
```
The parser would find the { in `'){�%P�` and try to parse from there, resulting in:
```
VM357:3264 Failed to parse main channel data: SyntaxError: Expected property name or '}' in JSON at position 1 (line 1 column 2)
at JSON.parse (<anonymous>)
at handleMainChannelEvent (<anonymous>:3243:29)
at RTCDataChannel.<anonymous> (<anonymous>:3279:9)
```

This prevents Dominant Speaker updates from working correctly.

### Solution
Modified the parsing logic to try multiple potential JSON start positions and continue searching when JSON parsing fails.

### DIFF
--- a/bots/teams_bot_adapter/teams_chromedriver_payload.js
+++ b/bots/teams_bot_adapter/teams_chromedriver_payload.js
@@ -1508,39 +1508,42 @@ const handleMainChannelEvent = (event) => {
     //realConsole?.log('handleMainChannelEvent jsonRawString', jsonRawString);
     
     // Find the start of the JSON data (looking for '[' or '{' character)
-    let jsonStart = 0;
     for (let i = 0; i < decodedData.length; i++) {
         if (decodedData[i] === 91 || decodedData[i] === 123) { // ASCII code for '[' or '{'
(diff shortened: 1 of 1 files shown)


noah-duncan:
Hey, just wanted to separate the decoding stuff into its own function. I did that in this PR: https://github.com/attendee-labs/attendee/pull/330

Could you review that and LMK if it looks good? It should be doing the exact same thing as your original PR.

wagao29:
Hi @noah-duncan , thanks for checking!
I completely agree that it's better to separate it into a function.
I left a minor comment that PR – please take a look. Everything else looks good.👍

PR #438, opened 2025-10-01
fix modal url if mode:debug for selfhost

berkantay:
Fix: Correct transcript API endpoint for self-hosted users

Previously, the transcript retrieval endpoint in the self-hosted option was incorrectly pointing to the managed API. This caused confusion and potential issues for users running Attendee in a self-hosted environment.

This PR updates the endpoint so that self-hosted deployments correctly reference the appropriate self-hosted API.


noah-duncan:
Hey @berkantay thanks for the contribution, this is a legimate issue. TBH, I think it would be cleaner to have a setting for the website domain, which can be http://localhost:8000, https://my.attendee.com, etc

PR #455, opened 2025-10-10
Add azure blob storage support

noah-duncan:
Azure Blob Storage is incompatible with S3, so we need to change our code to allow you to store files using it.

To enable Azure Storage:
1. Set the `STORAGE_PROTOCOL` env var to 'azure'
2. Set the `AZURE_RECORDING_STORAGE_CONTAINER_NAME` env var to the name of the blob storage container you want to store recordings in. This is similar to an s3 bucket.
3. Set either `AZURE_CONNECTION_STRING` or (`AZURE_ACCOUNT_KEY` AND `AZURE_ACCOUNT_NAME` env vars)
4. Set `AZURE_STORAGE_LINK_EXPIRATION_SECONDS` for how long presigned urls should last for. Default is 30 minutes. If you don't want the links to expire at all set `AZURE_STORAGE_USE_PERMANENT_LINKS` to 'true'

### DIFF
--- a/bots/tests/test_google_meet_bot_2.py
+++ b/bots/tests/test_google_meet_bot_2.py
@@ -5,7 +5,7 @@ from unittest.mock import MagicMock, call, patch
 
 import kubernetes
 from django.db import connection
-from django.test.testcases import TransactionTestCase
+from django.test.testcases import TransactionTestCase, override_settings
 from django.utils import timezone
 
 from bots.bot_adapter import BotAdapter
@@ -38,6 +38,43 @@ from bots.tests.mock_data import create_mock_file_uploader, create_mock_google_m
 from bots.web_bot_adapter.ui_methods import UiRetryableException
 
 
+@override_settings(
+    STORAGE_PROTOCOL="azure",
+    AZURE_RECORDING_STORAGE_CONTAINER_NAME="test-container",
+    CHARGE_CREDITS_FOR_BOTS=False,
+    STORAGES={  # build the exact structure your code expects
+        "default": {
+            "BACKEND": "storages.backends.azure_storage.AzureStorage",
+            "OPTIONS": {
+                "connection_string": "fake",
+                "account_key": "fake",
+                "account_name": "fake",
+                "expiration_secs": None,
+            },
+        },
+        "recordings": {
+            "BACKEND": "storages.backends.azure_storage.AzureStorage",
+            "OPTIONS": {
+                "connection_string": "fake",
+                "account_key": "fake",
+                "account_name": "fake",
+                "azure_container": "test-container",
+                "expiration_secs": None,
+            },
+        },
+        "bot_debug_screenshots": {
+            "BACKEND": "storages.backends.azure_storage.AzureStorage",
+            "OPTIONS": {
+                "connection_string": "fake",
+                "account_key": "fake",
+                "account_name": "fake",
+                "azure_container": "test-container",
+                "expiration_secs": None,
+            },
+        },
+        "staticfiles": {"BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage"},
+    },
+)
 class TestGoogleMeetBot2(TransactionTestCase):
     @classmethod
     def setUpClass(cls):


requirements.txt
```diff
@@ -90,4 +90,6 @@ kubernetes==32.0.0
 stripe==11.6.0
 tldextract==5.3.0
 aiortc==1.10.1
-deepgram-sdk==4.3.0
\ No newline at end of file
+deepgram-sdk==4.3.0
+azure-identity==1.25.1
+azure-storage-blob==12.26.0
```
pylead, line 95: It would be nicer to make s3/azure optional during installation / image build :see_no_evil: 
That could be additional `requirements_extras/s3.txt` & `requirements_extras/azure.txt`
That split would be better done later as a separate PR, because would require changes in CI and distribution.

Dockerfile
```diff
@@ -61,6 +61,12 @@ RUN apt-get install -y  pulseaudio pulseaudio-utils ffmpeg
 # Install Linux Kernel Dev
 RUN apt-get update && apt-get install -y linux-libc-dev
 
+# Update certificates
+RUN apt-get update && apt-get install -y \
+    ca-certificates \
+    && rm -rf /var/lib/apt/lists/* \
```
pylead, line 67: Just wandering: do we remove the repos to clean the space or for security reasons?

bots/bot_controller/azure_file_uploader.py
```diff
@@ -0,0 +1,91 @@
...
+            raise ValueError("Both 'container' and 'filename' are required")
+
+        # Prefer connection string if provided; otherwise fall back to account_name + account_key
+        if connection_string:
+            service_client = BlobServiceClient.from_connection_string(connection_string)
+        else:
+            if not (account_name and account_key):
+                raise ValueError("Provide either AZURE_CONNECTION_STRING or both AZURE_ACCOUNT_NAME and AZURE_ACCOUNT_KEY")
+            account_url = f"https://{account_name}.blob.core.windows.net"
+            service_client = BlobServiceClient(account_url=account_url, credential=account_key)
```
pylead, line 38: This might be easier to read :see_no_evil: 
```suggestion
        # Prefer connection string if provided; otherwise fall back to account_name + account_key
        if connection_string:
            service_client = BlobServiceClient.from_connection_string(connection_string)
        elif account_name and account_key:
            account_url = f"https://{account_name}.blob.core.windows.net"
            service_client = BlobServiceClient(account_url=account_url, credential=account_key)
        else:
            raise ValueError("Provide either AZURE_CONNECTION_STRING or both AZURE_ACCOUNT_NAME and AZURE_ACCOUNT_KEY")
```
I'm also wandering if we should expose settings names here:
- `AZURE_CONNECTION_STRING`
- `AZURE_ACCOUNT_NAME`
- `AZURE_ACCOUNT_KEY`

Since this class doesn't directly go into settings, it should probably mention constructor parameters here :see_no_evil: 
```python
raise ValueError("Provide either 'connection_string' or both 'account_name' and 'account_key'")
````

File: bots/bot_controller/bot_controller.py [hunk 4]
```diff
@@ -395,9 +397,9 @@ def upload_recording_to_external_media_storage_if_enabled(self):
 
         try:
             logger.info(f"Uploading recording to external media storage bucket {self.bot_in_db.external_media_storage_bucket_name()}")
-            file_uploader = FileUploader(
+            file_uploader = S3FileUploader(
```
pylead, line 400: This `upload_recording_to_external_media_storage_if_enabled` method seems to be strictly linked to S3 with current design. Shouldn't that be `get_file_uploader() called instead?
```suggestion
            file_uploader = self.get_file_uploader(
```

bots/bot_controller/bot_controller.py
```diff
@@ -409,6 +411,22 @@ def upload_recording_to_external_media_storage_if_enabled(self):
...
+                filename=self.get_recording_filename(),
+                connection_string=os.environ.get("AZURE_CONNECTION_STRING"),
+                account_key=os.environ.get("AZURE_ACCOUNT_KEY"),
+                account_name=os.environ.get("AZURE_ACCOUNT_NAME"),
+            )
+
+        return S3FileUploader(
+            bucket=os.environ.get("AWS_RECORDING_STORAGE_BUCKET_NAME"),
+            filename=self.get_recording_filename(),
+            endpoint_url=os.environ.get("AWS_ENDPOINT_URL"),
```
pylead: Shouldn't that variables be taken from settings instead of directly from `os.environ`? :see_no_evil:
The django `settings` could work as a one source of truth here.

bots/models.py
```diff
@@ -1525,11 +1525,19 @@ class TranscriptionProviders(models.IntegerChoices):
     ELEVENLABS = 7, "ElevenLabs"
 
 
+from storages.backends.azure_storage import AzureStorage
 from storages.backends.s3boto3 import S3Boto3Storage
 
 
-class RecordingStorage(S3Boto3Storage):
-    bucket_name = settings.AWS_RECORDING_STORAGE_BUCKET_NAME
+class RecordingStorage(AzureStorage if settings.STORAGE_PROTOCOL == "azure" else S3Boto3Storage):
```
pylead: That could fail if model is imported directly before settings init is done. :see_no_evil:

bots/models.py
```diff
@@ -1525,11 +1525,19 @@ class TranscriptionProviders(models.IntegerChoices):
...
+    if settings.STORAGE_PROTOCOL == "azure":
+        azure_container = settings.AZURE_RECORDING_STORAGE_CONTAINER_NAME
+        account_name = settings.DEFAULT_STORAGE_BACKEND.get("OPTIONS").get("account_name")
+        account_key = settings.DEFAULT_STORAGE_BACKEND.get("OPTIONS").get("account_key")
+        connection_string = settings.DEFAULT_STORAGE_BACKEND.get("OPTIONS").get("connection_string")
+        expiration_secs = settings.DEFAULT_STORAGE_BACKEND.get("OPTIONS").get("expiration_secs")
```
pylead: I think it would crash without `OPTIONS`, we would need something like this:
```suggestion
        account_name = settings.DEFAULT_STORAGE_BACKEND.get("OPTIONS", {}).get("account_name")
        account_key = settings.DEFAULT_STORAGE_BACKEND.get("OPTIONS", {}).get("account_key")
        connection_string = settings.DEFAULT_STORAGE_BACKEND.get("OPTIONS", {}).get("connection_string")
        expiration_secs = settings.DEFAULT_STORAGE_BACKEND.get("OPTIONS", {}).get("expiration_secs")
```

pylead: Maybe we should define that settings reading directly in the `AzureStorage` / `S3Boto3Storage` classes? Or do you think it's the bridge class `RecordingStorage` responsibility to init from settings? :see_no_evil:

bots/models.py
```diff
@@ -1569,6 +1577,10 @@ def __str__(self):
     def url(self):
         if not self.file.name:
             return None
+
+        if settings.STORAGE_PROTOCOL == "azure":
+            return self.file.url
```
pylead, line 1578: Maybe it would be nicer to define `self.file.url` abstraction for S3 storage  instead of calling S3-related code below? :see_no_evil:

Dockerfile
jpabl0m: Security

requirements.txt
noah-duncan: Yea sounds good! Feel free to think about how best to do that.

bots/bot_controller/azure_file_uploader.py
noah-duncan: Thanks, makes sense!

bots/bot_controller/bot_controller.py
noah-duncan: As discussed in our slack thread, we'll leave this for a latter PR. Issue here: https://github.com/attendee-labs/attendee/issues/459

bots/bot_controller/bot_controller.py
noah-duncan: Yea good call out

bots/models.py
noah-duncan: It would probably be OK to do that and definitely would be cleaner, but the S3 stuff has been finicky in the past and there's a 5% chance that would change the actual operation and break something for someone. Would need to look into the django storages code and to see if there are differences between the abstraction and the current code.

bots/models.py
noah-duncan: Lol true. That code always looked kinda ugly, just refactored to do it a cleaner way.

PR #466, opened 2025-10-15
feat: add Kyutai streaming transcriber with semantic VAD

docteurZ:
- Added KyutaiStreamingTranscriber for real-time STT 
- Implemented buffered audio streaming
- Integrated semantic VAD for natural utterance segmentation

### DIFF
--- a/bots/bot_controller/per_participant_streaming_audio_input_manager.py
+++ b/bots/bot_controller/per_participant_streaming_audio_input_manager.py
@@ -139,7 +139,6 @@ class PerParticipantStreamingAudioInputManager:
                 sample_rate=self.sample_rate,
                 metadata=metadata,
                 interim_results=True,
-                model=self.bot.transcription_settings.kyutai_model(),
                 api_key=self.kyutai_api_key,
                 save_utterance_callback=kyutai_save_utterance_callback,
             )
--- a/bots/models.py
+++ b/bots/models.py
@@ -614,9 +614,6 @@ class TranscriptionSettings:
     def deepgram_replace_settings(self):
         return self._settings.get("deepgram", {}).get("replace", [])
 
-    def kyutai_model(self):
-        return self._settings.get("kyutai", {}).get("model", None)
-
     def kyutai_server_url(self):
         return self._settings.get("kyutai", {}).get("server_url", None)
 
--- a/bots/transcription_providers/kyutai/kyutai_streaming_transcriber.py
+++ b/bots/transcription_providers/kyutai/kyutai_streaming_transcriber.py
@@ -122,7 +122,6 @@ class KyutaiStreamingTranscriber:
         sample_rate,
         metadata=None,
         interim_results=True,
-        model=None,
         api_key=None,
         save_utterance_callback=None,
         max_retry_time=300,
@@ -137,7 +136,6 @@ class KyutaiStreamingTranscriber:
             sample_rate: Audio sample rate (Kyutai uses 24000 Hz)
             metadata: Optional metadata to send with the connection
             interim_results: Whether to receive interim results
-            model: Model name to use (optional)
             api_key: API key for authentication (optional)
             save_utterance_callback: Callback function for saving utterances
                       (receives transcript text)
@@ -150,7 +148,6 @@ class KyutaiStreamingTranscriber:
         self.sample_rate = sample_rate
         self.metadata = metadata or {}
         self.interim_results = interim_results
-        self.model = model
         self.api_key = api_key
         self.save_utterance_callback = save_utterance_callback
         self.max_retry_time = max_retry_time


bots/tasks/process_utterance_task.py
```diff
@@ -38,6 +39,9 @@ def get_transcription(utterance):
             transcription, failure_data = get_transcription_via_sarvam(utterance)
         elif utterance.transcription_provider == TranscriptionProviders.ELEVENLABS:
             transcription, failure_data = get_transcription_via_elevenlabs(utterance)
+        # Streaming-only providers that don't support post-processing in this task (could be done though)
+        elif utterance.transcription_provider == TranscriptionProviders.KYUTAI:
+            return None, {"reason": TranscriptionFailureReasons.STREAMING_ONLY_PROVIDER, "error": ("Kyutai is a streaming-only transcription provider. Transcriptions are generated in real-time during the bot session")}
```
noah-duncan: If we ever hit this case, it means something is wrong with the code. Not sure if we need this.  Maybe just change this line `raise Exception(f"Unknown transcription provider: {utterance.transcription_provider}")` or `raise Exception(f"Unknown or streaming-only transcription provider: {utterance.transcription_provider}")`

dev.docker-compose.yaml
```diff
@@ -34,7 +34,7 @@ services:
     networks:
       - attendee_network
     ports:
-      - "8000:8000"
+      - "8002:8000"
```
noah-duncan: Did you mean to commit this?

requirements.txt
```diff
@@ -91,7 +91,8 @@ stripe==11.6.0
 tldextract==5.3.0
 aiortc==1.10.1
 deepgram-sdk==4.3.0
+msgpack==1.1.0
```
noah-duncan, line 94: do we need this for somethign?

bots/bot_controller/bot_controller.py
```diff
@@ -82,7 +82,10 @@ class BotController:
     UTTERANCE_TERMINATION_WAIT_TIME_SECONDS = 300
 
     def per_participant_audio_input_manager(self):
-        if self.bot_in_db.transcription_settings.deepgram_use_streaming():
+        # Use streaming manager for Deepgram streaming or Kyutai
+        # (Kyutai is streaming-only)
+        provider = self.get_recording_transcription_provider()
+        if self.bot_in_db.transcription_settings.deepgram_use_streaming() or provider == TranscriptionProviders.KYUTAI:
```
noah-duncan: Would look cleaner if the `self.bot_in_db.transcription_settings.deepgram_use_streaming() or provider == TranscriptionProviders.KYUTAI` logic was in a `use_streaming` method on transcription_settings

bots/bot_controller/per_participant_streaming_audio_input_manager.py
```diff
@@ -1,35 +1,68 @@
...
+        # Normalize by max possible value for 16-bit audio (32768)
+        return rms / 32768
+    except (ValueError, TypeError, BufferError, FloatingPointError):
+        # If there's any issue with the audio data, treat as silence
+        return 0.0
 
 
 class PerParticipantStreamingAudioInputManager:
     def __init__(self, *, get_participant_callback, sample_rate, transcription_provider, bot):
-        self.queue = queue.Queue()
```
noah-duncan, line 23: thanks for removing

bots/bot_controller/per_participant_streaming_audio_input_manager.py
```diff
@@ -1,35 +1,68 @@
...
-
         self.get_participant_callback = get_participant_callback
 
         self.utterances = {}
         self.sample_rate = sample_rate
 
         self.last_nonsilent_audio_time = {}
 
-        self.SILENCE_DURATION_LIMIT = 10  # seconds
+        self.SILENCE_DURATION_LIMIT = 300  # 5 minutes of inactivity
```
noah-duncan: I think this was set more aggressively for deepgram because they have fairly tight rate limits on # of concurrent stream connections. Could you make this conditional?

bots/bot_controller/per_participant_streaming_audio_input_manager.py
```diff
@@ -53,61 +90,142 @@ def get_deepgram_api_key(self):
...
+        return server_url, api_key
+
     def create_streaming_transcriber(self, speaker_id, metadata):
-        logger.info(f"Creating streaming transcriber for speaker {speaker_id}")
         if self.transcription_provider == TranscriptionProviders.DEEPGRAM:
             metadata_list = [f"{key}:{value}" for key, value in metadata.items()] if metadata else None
             return DeepgramStreamingTranscriber(
                 deepgram_api_key=self.deepgram_api_key,
                 interim_results=True,
-                language=self.bot.transcription_settings.deepgram_language(),
```
noah-duncan, line 63: You need to keep these, right?

bots/bot_controller/per_participant_streaming_audio_input_manager.py
```diff
@@ -53,61 +90,142 @@ def get_deepgram_api_key(self):
...
-                model=self.bot.transcription_settings.deepgram_model(),
-                callback=self.bot.transcription_settings.deepgram_callback(),
                 sample_rate=self.sample_rate,
+                callback=self.utterance_handler.handle_utterance,
```
noah-duncan: To preserve the existing behavior, can you put `callback=self.bot.transcription_settings.deepgram_callback()` back? The existing behavior is that deepgram streaming is only supported IF you specify a callback url, so deepgram talks directly to your server and bypasses Attendee. This change removed that.

In the long term, we definitely want to allow for both options.

noah-duncan (changes requested):

dev.docker-compose.yaml
docteurZ: mb, sorry about that.

requirements.txt
docteurZ: Added msgpack to requirements because the Kyutai Rust server uses MessagePack binary serialization for communication. This is necessary for the Python client to encode/decode messages when interfacing with the Kyutai streaming transcription service.

noah-duncan: Ok, no problem.

bots/bot_controller/per_participant_streaming_audio_input_manager.py
docteurZ: mb, good catch. sorry about that

bots/bot_controller/per_participant_streaming_audio_input_manager.py
docteurZ: it make sense! I actually forgot to restore that behavior when I was testing my changes. thx

bots/bot_controller/per_participant_streaming_audio_input_manager.py
```diff
@@ -53,61 +95,143 @@ def get_deepgram_api_key(self):
...
+
     def create_streaming_transcriber(self, speaker_id, metadata):
-        logger.info(f"Creating streaming transcriber for speaker {speaker_id}")
         if self.transcription_provider == TranscriptionProviders.DEEPGRAM:
             metadata_list = [f"{key}:{value}" for key, value in metadata.items()] if metadata else None
             return DeepgramStreamingTranscriber(
                 deepgram_api_key=self.deepgram_api_key,
                 interim_results=True,
                 language=self.bot.transcription_settings.deepgram_language(),
-                model=self.bot.transcription_settings.deepgram_model(),
```
noah-duncan, line 64: You shouldn't be removing this attribute, right?

PR #468, opened 2025-10-16
feat: JSON format logs

nijave:
Main use case is making "log level" a field so Datadog or log aggregator can parse and filter correctly

Not sure if it'd be better to pull log config into `base.py`

Example
```
attendee-scheduler-local-1  | {"timestamp": "2025-10-16T22:09:31.572540+00:00", "name": "bots.management.commands.run_scheduler", "levelname": "INFO", "message": "Scheduler daemon exited"}
```

### DIFF
--- a/attendee/settings/production.py
+++ b/attendee/settings/production.py
@@ -1,4 +1,5 @@
 import os
+import sys
 
 import dj_database_url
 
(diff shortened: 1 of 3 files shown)


attendee/settings/production.py
```diff
@@ -51,9 +53,15 @@
...
     "version": 1,
     "disable_existing_loggers": False,
+    "formatters": {
+        "plain": {"format": "{levelname} {message}", "style": "{"},
+        "json": {"class": "attendee.logging.ISOJsonFormatter", "format": "%(timestamp)s %(name)s %(levelname)s %(message)s"},
+    },
     "handlers": {
         "console": {
             "class": "logging.StreamHandler",
+            "stream": sys.stdout,
```
noah-duncan, line 63: Nice! It was always slightly annoying seeing the logs show in google cloud as errors.

noah-duncan (approved):
Thanks @nijave ! Good job at making the log format optional while making two other changes (setting CELERY_WORKER_HIJACK_ROOT_LOGGER to false and directing the stream to stdout instead of stderror) that should be applied everywhere.

PR #469, opened 2025-10-16
Fix GCP Cloud Logging severity classification with JSON stdout logging

docteurZ:
Previously, logs were being written to stderr using Python's default StreamHandler, causing GCP Cloud Logging to incorrectly classify all logs as ERROR severity, regardless of their actual level (INFO, WARNING, DEBUG, etc.). 

Note: This currently only applies to worker pods. Web/API pods will need separate configuration to enable proper JSON logging.

### DIFF
--- a/attendee/settings/production-gke.py
+++ b/attendee/settings/production-gke.py
@@ -1,3 +1,4 @@
+import logging
 import os
 
 import dj_database_url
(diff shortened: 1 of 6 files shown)


nijave:
I also have https://github.com/attendee-labs/attendee/pull/468 which I think should update everything. You need to set `ATTENDEE_LOG_FORMAT=json` to apply the JSON config

PR #503, opened 2025-11-13
fix deliver_webhook task emitting before transaction commit 

chawlamark:
Ensure deliver_webhook task reads committed WebhookDeliveryAttempt

- Wrap task trigger in `transaction.on_commit` to avoid DoesNotExist errors
  when the row hasn’t been committed yet.

getting a lot of does not exist 
Screenshot 2025-11-13 at 4 42 34 PM

### DIFF
--- a/bots/webhook_utils.py
+++ b/bots/webhook_utils.py
@@ -4,6 +4,7 @@ import hmac
 import json
 import logging
 import uuid
+from functools import partial
 
 from django.db import transaction
 
@@ -56,7 +57,7 @@ def trigger_webhook(webhook_trigger_type, bot=None, calendar=None, zoom_oauth_co
 
         from bots.tasks.deliver_webhook_task import deliver_webhook
 
-        transaction.on_commit(lambda: deliver_webhook.delay(delivery_attempt.id))
+        transaction.on_commit(partial(deliver_webhook.delay, delivery_attempt.id))
 
     return len(delivery_attempts)
 


noah-duncan (changes requested):
Hey @chawlamark thanks for the PR! Curious how often you're seeing this and if it indicates a deeper issue.

There's actually a subtle bug that will happen in the case for multiple webhook subscriptions. It's going to use the *last* delivery_attempt.id value in the loop.

You can verify by adding this test to your code in `bots/tests/test_webhooks.py`

```
    @patch("bots.tasks.deliver_webhook_task.deliver_webhook")
    def test_trigger_webhook_uses_distinct_attempt_ids_for_multiple_subscriptions(self, mock_deliver):
        """
        Triggering webhooks for multiple subscriptions should schedule tasks
        with distinct WebhookDeliveryAttempt IDs.

        With the buggy lambda (lambda: deliver_webhook.delay(delivery_attempt.id)),
        both on_commit callbacks end up using the last delivery_attempt.id.
        """

        from bots.webhook_utils import trigger_webhook

        # Make sure we have two project-level webhook subscriptions for the same trigger
        WebhookSubscription.objects.all().delete()
        WebhookSubscription.objects.create(
            project=self.project,
            url="https://example.com/webhook1",
            triggers=[WebhookTriggerTypes.BOT_STATE_CHANGE],
        )
        WebhookSubscription.objects.create(
            project=self.project,
            url="https://example.com/webhook2",
            triggers=[WebhookTriggerTypes.BOT_STATE_CHANGE],
        )

        # Clear any existing delivery attempts
        WebhookDeliveryAttempt.objects.all().delete()

        # Run inside an explicit transaction so transaction.on_commit defers execution
        with transaction.atomic():
            trigger_webhook(
                webhook_trigger_type=WebhookTriggerTypes.BOT_STATE_CHANGE,
                bot=self.bot,
                payload={"test": "closure_bug"},
            )

        # After the transaction commits, on_commit callbacks should have fired
        self.assertEqual(WebhookDeliveryAttempt.objects.count(), 2)
        self.assertEqual(mock_deliver.delay.call_count, 2)

        # IDs that were actually passed to deliver_webhook.delay(...)
        attempt_ids_called = [call.args[0] for call in mock_deliver.delay.call_args_list]

        # IDs of the delivery attempts we created
        attempt_ids_in_db = list(WebhookDeliveryAttempt.objects.values_list("id", flat=True))

        # With correct code, the sets match (two distinct IDs).
        # With the buggy lambda, attempt_ids_called will contain the same ID twice.
        self.assertEqual(set(attempt_ids_called), set(attempt_ids_in_db))

```

If you run the test with `DJANGO_SETTINGS_MODULE=attendee.settings.test python manage.py test --keepdb bots.tests.test_webhooks.WebhookDeliveryTest.test_trigger_webhook_uses_distinct_attempt_ids_for_multiple_subscriptions` it'll fail.

Using functools.partial instead of lambda gets the tests to pass though.

Image

Relevant article: https://medium.com/skiller-whale/late-binding-variables-its-a-trap-c17af980164f

chawlamark:
@noah-duncan thank you for catching that, I appreciate it, I made the change you suggested. Also, it happened about 300+ times in the last week in our production env but this is a common issue we see with our infrastructure/stack even in other django apps

PR #510, opened 2025-11-18
Ability to custom pod k8s configuration via env vars

noah-duncan:
Adds ability to customize bot pod k8s configuration via env vars. This allows DevOps to change the pod configuration without editing the application code. It uses JSON patch semantics to support open-ended changes.

A bot pod k8s configuration is stored in an env var with the name BOT_POD_SPEC_[configuration type]. The configuration is represented by a list of JSON patch operations that act on the default bot pod spec.

Currently, two configuration types are supported - default and scheduled.

The configuration type for a bot is chosen as follows
If the bot's join_at is more than `SCHEDULED_BOT_POD_SPEC_MARGIN_SECONDS` into the future and a `scheduled` configuration type is defined, then use the `scheduled` configuration type. Otherwise use the `default` configuration type.

In the future (if needed), additional configuration names can be passed in via the bot api call which will override the logic above.

Example env var values

```
BOT_POD_SPEC_DEFAULT='[{"op":"add","path":"/metadata/labels/default-bot","value":"true"}]'
BOT_POD_SPEC_SCHEDULED='[{"op":"add","path":"/metadata/labels/scheduled-bot","value":"true"}]'
```

### DIFF
--- a/attendee/settings/base.py
+++ b/attendee/settings/base.py
@@ -180,6 +180,7 @@ REST_FRAMEWORK = {
     },
 }
 
+DISABLE_RATE_LIMITING = os.getenv("DISABLE_RATE_LIMITING", "false") == "true"
 SPECTACULAR_SETTINGS = {
     "TITLE": "Attendee API",
     "DESCRIPTION": "Meetings bots made easy",
--- a/bots/bot_controller/bot_controller.py
+++ b/bots/bot_controller/bot_controller.py
@@ -811,7 +811,7 @@ class BotController:
                     message = self.pubsub.get_message(timeout=1.0)
                     if message:
                         # Schedule Redis message handling in the main GLib loop
-                        GLib.idle_add(lambda: self.handle_redis_message(message))
+                        GLib.idle_add(self.handle_redis_message, message)
                 except Exception as e:
                     # If this is a certain type of exception, we can attempt to reconnect
                     if isinstance(e, redis.exceptions.ConnectionError) and "Connection closed by server." in str(e):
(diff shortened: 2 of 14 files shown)


bots/models.py
```diff
@@ -706,6 +708,12 @@ def set_heartbeat(self):
                     raise
                 continue
 
+    def bot_pod_spec_type(self) -> BotPodSpecType:
+        # If join_at is greater than SCHEDULED_BOT_POD_SPEC_MARGIN_SECONDS seconds into the future, use the scheduled pod spec
+        if self.join_at and self.join_at - timedelta(seconds=os.getenv("SCHEDULED_BOT_POD_SPEC_MARGIN_SECONDS", 120)) > timezone.now():
+            return BotPodSpecType.SCHEDULED
+        return BotPodSpecType.DEFAULT
```
pylead: Maybe making this a property would be more pythonic :see_no_evil: 
e.g. it could be used like `bot.pod_spec_type`

bots/bot_pod_creator/bot_pod_creator.py
```diff
@@ -298,6 +346,8 @@ def create_bot_pod(
             )
         )
 
+        bot_pod_dict = self.apply_spec_to_bot_pod(bot_pod)
```
pylead: The `bot_pod_dict` name can be confusing. It can be misinterpreted like it were containing a map of `{"bot_id":"pod_id"}` or something like that.

Maybe `bot_pod_spec_data` makes it clearer? :see_no_evil:

pylead (approved):
So you finally decided to keep ad-hoc (default) / scheduled logic inside attendee server.
Thank you :pray:

noah-duncan:
> Seems reasonable. Only thing is whether it's possible to load and validate patches ahead of time (in settings)? That'd generate errors faster and you could then unmarshal json patches a single time

Yea that makes sense. Will add in a follow-up PR. Merging now so others can use.

PR #512, opened 2025-11-18
fix: Auto-recreate Kyutai transcriber on connection failure

docteurZ:
Raises ConnectionError when WebSocket connection fails, allowing manager to delete and recreate transcriber on next audio chunk.

Added  flag to avoid false positives during initial connection. During reconnection attempts, audio is silently dropped until connection succeeds or transcriber is recreated.

### DIFF
--- a/bots/bot_controller/per_participant_streaming_audio_input_manager.py
+++ b/bots/bot_controller/per_participant_streaming_audio_input_manager.py
@@ -146,11 +146,18 @@ class PerParticipantStreamingAudioInputManager:
             raise Exception(f"Unsupported transcription provider: {self.transcription_provider}")
 
     def find_or_create_streaming_transcriber_for_speaker(self, speaker_id):
-        if speaker_id not in self.streaming_transcribers:
-            metadata = {"bot_id": self.bot.object_id, **(self.bot.metadata or {}), **self.get_participant_callback(speaker_id)}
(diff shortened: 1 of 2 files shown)


bots/bot_controller/per_participant_streaming_audio_input_manager.py
```diff
@@ -173,12 +180,19 @@ def add_chunk(self, speaker_id, chunk_time, chunk_bytes):
...
+            # Create transcriber if needed
             streaming_transcriber = self.find_or_create_streaming_transcriber_for_speaker(speaker_id)
             if streaming_transcriber:
-                streaming_transcriber.send(chunk_bytes)
-            else:
-                logger.warning(f"Failed to create transcriber for speaker {speaker_id}")
+                # Send audio
+                try:
+                    streaming_transcriber.send(chunk_bytes)
+                except Exception as e:
```
noah-duncan, line 189: Would it make sense to catch only a connection failure exception instead catching any exception?

docteurZ: Most likely, yes. I started broad intentionally and plan to narrow it down as we identify the specific exceptions we need to handle

noah-duncan: Ok that's fair.

kylefossum (approved):
Sorry for the delay, we'll tweak the branch protection rules

PR #518, opened 2025-11-19
Feature: Support mariadb and MinIO

ready4god2513:
At Teamwork most of our stack (especially in our development environment) uses minio for our blob storage and mariadb for our data store.  I've provided options here to run attendee with either aws credentials or minio (which shares the same API interface as aws s3), as well as postgres or mariadb/mysql for the database.

I ended up having to split out some of the migrations (and change the order as constraints were defined before schema) where constraint support was not available in mariadb.

I have run docker-compose up for both postgres as well as mariadb support, ensured migrations ran from scratch in both systems.  Everything is running for me in development as expected with both database systems.

image

I have tested this on MacOS.

### DIFF
--- a/dev.docker-compose.yaml
+++ b/dev.docker-compose.yaml
@@ -22,7 +22,7 @@ services:
       - AWS_RECORDING_STORAGE_BUCKET_NAME=${AWS_RECORDING_STORAGE_BUCKET_NAME:-}
     command: "celery -A attendee worker -l INFO"
     security_opt:
-      - seccomp=unconfined
+      - seccomp=${PWD}/bots/web_bot_adapter/chrome_seccomp.json
 
   attendee-scheduler-local:
     build: ./


noah-duncan:
Hi @ready4god2513 , thanks for the PR. I appreciate you confirming that the code works with both Postgres and MariaDB. 

Still, supporting both Postgres and MariaDB does significantly increase the complexity of this project and it may constrain what we can do in the future. So would like to see if there's an alternative that avoids this. I'd love to chat and learn more about how TeamWork is planning to use Attendee. 

Would you mind either emailing me at noah@attendee.dev, joining our [slack community](https://join.slack.com/t/attendeecommu-rff8300/shared_invite/zt-2uhpam6p2-ZzLAoVrljbL2UEjqdSHrgQ) or [booking a meeting](https://calendly.com/noah-attendee/30min)?

PR #521, opened 2025-11-20
Custom Async

dvillaume:
Hi,

This PR aims to use **self-hosted transcription platform** in a generic way.
It assumes the self-hosted platforms respects some common rules.

Tell me if this can be usefull or if i need to adjust anything.


noah-duncan:
Hi @dvillaume thanks for the PR. I just want to make sure that this capability cannot do what you need:
https://docs.attendee.dev/guides/transcription#transcription__choosing-the-right-transcription-provider__openai__note-on-custom-openai-proxy-servers

We have the ability specific a custom server via the OPENAI_BASE_URL and OPENAI_MODEL_NAME env vars.

PR #537, opened 2025-11-26
Add gpt 4o transcribe diarize

ryusock:
Adds new OpenAI model `gpt-4o-transcribe-diarize`, as requested in [this issue](https://github.com/attendee-labs/attendee/issues/471)

The model [accepts a new `response_format` value](https://platform.openai.com/docs/api-reference/audio/createTranscription#audio_createtranscription-response_format) `diarized_json`. Since this is the information users will want from the model, this will be the default response format, instead of `json` for the other OpenAI models.

The model also is most useful when it also [accepts the parameter `chunking_strategy`](https://platform.openai.com/docs/api-reference/audio/createTranscription#audio_createtranscription-chunking_strategy), which accepts either "auto" or an object that is formatted as a Server VAD object. This PR adds supports for this parameter because OpenAI says for the diarized model, this parameter is [required when the input is longer than 30 seconds](https://platform.openai.com/docs/guides/speech-to-text#speaker-diarization). 

**Note**
The body of the diarized response from OpenAI is transformed to Attendee's schema. The schema isn't explicitly defined, which will come in a quick follow-up. But we don't want to break usage of the current `transcription` object. OpenAI's response sends back chunks by `segment`, which are identified by the slight misnomer `word` in the Attendee response.

### DIFF
--- a/bots/tasks/process_utterance_task.py
+++ b/bots/tasks/process_utterance_task.py
@@ -29,7 +29,7 @@ def transform_diarized_json_to_schema(result):
         segment_text = segment.get("text", "")
         segment_start = segment.get("start", 0.0)
         segment_end = segment.get("end", segment_start)
-        speaker = segment.get("speaker", "")
+        speaker = segment.get("speaker", None)
 
         word_obj = {
             "word": segment_text,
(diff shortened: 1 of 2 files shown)


bots/tasks/process_utterance_task.py
```diff
@@ -14,6 +14,37 @@
...
+    # Extract segments (OpenAI sends each "word" as a separate segment, may contain multiple words).
+    # We will transform each segment into a word object, despite the fact that it may contain multiple words.
+    segments = result.get("segments", [])
+    words = []
+
+    for segment in segments:
+        segment_text = segment.get("text", "")
+        segment_start = segment.get("start", 0.0)
+        segment_end = segment.get("end", segment_start)
+        speaker = segment.get("speaker", "")
```
noah-duncan: In the other transcription provider that supports the speaker attribute we passed 'None' instead of "" when no speaker was assigned.

PR #561, opened 2025-12-12
Fix: Convert custom_async nested objects to JSON in multipart form data

dvillaume:
- Convert dict/list values in custom_async to JSON string using json.dumps()
- Convert PCM audio to MP3 before sending (consistent with other providers)
- Update tests to reflect MP3 conversion and test JSON serialization


noah-duncan:
Hey @dvillaume looks good, you just need to run `ruff check --fix; ruff format` on your code to fix the linting issue, that's causing CI to fail. Also, I made you a collaborator of the repo, so you don't need to fork when you want to contribute.

PR #568, opened 2025-12-15
Add bot log entity and webhook for when auto CC is not enabled

ryusock:


bots/models.py
```diff
@@ -1775,6 +1775,77 @@ def create_event(
...
+    def type_to_api_code(cls, value):
+        """Returns the API code for a given type value"""
+        mapping = {
+            cls.UNCATEGORIZED: "uncategorized",
+            cls.CLOSED_CAPTIONS_DISABLED: "closed_captions_disabled",
+        }
+        return mapping.get(value)
+
+
+class Log(models.Model):
```
noah-duncan: I'm worried this name will make people think its some generic log class. I'd call it BotLogEvent, to mmake it more clear this is something different

bots/web_bot_adapter/web_bot_adapter.py
```diff
@@ -851,6 +853,16 @@ def ready_to_show_bot_image(self):
         self.send_message_callback({"message": self.Messages.READY_TO_SHOW_BOT_IMAGE})
 
     def could_not_enable_closed_captions(self):
+        if getattr(self, "bot", None) is not None:
+            BotLogManager.create_bot_log(
+                bot=self.bot,
```
ryusock: It feels wrong to have to feed the bot instance here, in the adapter (need the bot to create the BotLog). Not only that, importing the BotLogEventType class from models.py I think creates a circular import because this adapter can't import it.

Based on the pattern I'm seeing in the rest of the class, anything to do with the bot should be invoking a callback method from the bot_controller class, the `take_action_based_on_message_from_adapter` method. What if we invoked the callback with `send_message_callback`, and then created the BotLog in bot_controller?

noah-duncan: Yep, the adapter should pass a message to bot controller and you should create it there. The adapters are designed to be loosely coupled to the rest of the system, so they shouldn't interact directly w the database.

ryusock: Thanks, just wanted to confirm. I'll do it that way!

bots/bot_adapter.py
```diff
@@ -24,6 +24,8 @@ class Messages:
         JOINING_BREAKOUT_ROOM = "Joining breakout room"
         LEAVING_BREAKOUT_ROOM = "Leaving breakout room"
         BOT_RECORDING_PERMISSION_DENIED = "Bot recording permission denied"
+        # Bot log event, does not indicate bot state change
```
noah-duncan: FYI, there are a few other things in this enum which aren't state changes, such as `READY_TO_SHOW_BOT_IMAGE`.

bots/models.py
```diff
@@ -2725,6 +2799,7 @@ def _get_mapping(cls):
             cls.CALENDAR_STATE_CHANGE: "calendar.state_change",
             cls.ASYNC_TRANSCRIPTION_STATE_CHANGE: "async_transcription.state_change",
             cls.ZOOM_OAUTH_CONNECTION_STATE_CHANGE: "zoom_oauth_connection.state_change",
+            cls.LOG_MESSAGE: "log.message",
```
noah-duncan: This doesn't really follow the naming convention of the trigger types. The others have the format `<object>:<verb>`. I think you'd call this `bot_logs.update`

bots/models.py
```diff
@@ -1775,6 +1775,79 @@ def create_event(
...
+            cls.UNCATEGORIZED: "uncategorized",
+            cls.COULD_NOT_ENABLE_CLOSED_CAPTIONS: "could_not_enable_closed_captions",
+        }
+        return mapping.get(value)
+
+
+class BotLog(models.Model):
+    bot = models.ForeignKey(Bot, on_delete=models.CASCADE, related_name="logs")
+    level = models.IntegerField(choices=BotLogLevels.choices, default=BotLogLevels.INFO, null=False)
+    event_type = models.IntegerField(choices=BotLogEventTypes.choices, default=BotLogEventTypes.UNCATEGORIZED, null=False)
```
noah-duncan: To be consistent with the naming for BotEvent, this should be named `log_type` and BotLogEventTypes should be named BotLogTypes

bots/models.py
```diff
@@ -1775,6 +1775,79 @@ def create_event(
...
+    def type_to_api_code(cls, value):
+        """Returns the API code for a given type value"""
+        mapping = {
+            cls.UNCATEGORIZED: "uncategorized",
+            cls.COULD_NOT_ENABLE_CLOSED_CAPTIONS: "could_not_enable_closed_captions",
+        }
+        return mapping.get(value)
+
+
+class BotLog(models.Model):
```
noah-duncan: IDK if this is worth it, but it would clarify the name a bit if we named it BotLogItem or BotLogEntry. BC BotLog kind of makes it sound a single instance of this class represents an entire collection of logs. Up to you.

ryusock:
All the suggestions make sense. I'll ping you once those names and docs are updated

Issue #165, opened 2025-04-10
Multilingual support


3abassi:
Hello, 

Thank you for the wonderful repo!

i came across this comment in your source code:

        # nova-3 does not have multilingual support yet, so we need to use nova-2 if we're transcribing with a non-default language under ~/attendee\bots\tasks\process_utterance_task.py

I went throught the Deepgram docs and they say the following: 

To enable Multilingual Codeswitching, use the following language parameter in the query string when you call Deepgram’s /listen endpoint :
language=multi, they also provide an example using nova-3 (https://developers.deepgram.com/docs/language-detection)

what do you think?

noah-duncan:
Hi @3abassi , yea if nova-3 now supports multi-lingual support, then we should start using it. Currently, focusing on other issues so I will get to this in a week or two unless you or someone else wants to add it.

noah-duncan:
Hi @3abassi , thanks for pointing that out. If nova-3 now supports multi-lingual support, then we should definitely start using it. Currently, I am focusing on other issues so I will get to this in a week or two unless you or someone else wants to add it.

noah-duncan:
We support this now

PR #590, opened 2026-01-03
feat(transcription): use Nova-3 as default Deepgram model

sgaluza:
## Summary
Nova-3 now supports all languages, remove nova-2 fallback logic.

## Why this change
The original code fell back to Nova-2 for non-English languages because Nova-3 didn't support them at launch. Since November 2024, Nova-3 supports 50+ languages including Russian, Ukrainian, Hindi, etc.

Nova-2 and Nova-3 have the same pricing ($0.0043/min for batch), but Nova-3 offers better accuracy and noise handling.

## Testing
Verified via Deepgram API that Nova-3 accepts:
- language="ru" (Russian)
- language="uk" (Ukrainian)  
- language="hi" (Hindi)
- language="multi" (auto-detect)
- detect_language=true

All requests returned valid transcription results without errors.

## References
- Related to #165
- Supersedes #191

### DIFF
--- a/bots/tests/test_process_utterance_task.py
+++ b/bots/tests/test_process_utterance_task.py
@@ -201,6 +201,74 @@ class BotModelRedactionSettingsTest(TransactionTestCase):
         self.assertEqual(result1, ["pii", "pci"])
 
 
+class BotModelDeepgramModelTest(TransactionTestCase):
+    """Unit tests for Bot model deepgram_model method."""
+
+    def setUp(self):
+        self.organization = Organization.objects.create(name="Test Org")
+        self.project = Project.objects.create(name="Test Project", organization=self.organization)
+
+    def _create_bot_with_settings(self, settings):
+        """Helper to create a bot with specific settings."""
+        return Bot.objects.create(project=self.project, meeting_url="https://zoom.us/j/test", settings=settings)
+
+    def test_deepgram_model_returns_nova3_by_default(self):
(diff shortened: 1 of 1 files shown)


noah-duncan (changes requested):
Hi @sgaluza thanks for the PR and thanks for testing the various configurations to confirm that nova-3 will work. 

In order to maintain full backwards compatibility for existing codebases that call the Attendee API, I think we will need to set the model to nova-2 if `self.deepgram_langage()` is any of the following
```
Mandarin (Simplified): zh, zh-CN, zh-Hans

Mandarin (Traditional): zh-TW, zh-Hant

Cantonese (Traditional): zh-HK 

Thai: th, th-TH
```

According to the docs [here](https://developers.deepgram.com/docs/models-languages-overview), these languages are supported by nova-2 but not yet by nova-3.

sgaluza (commented):
Hi @noah-duncan, thanks for the review!

I've added the fallback logic to use `nova-2` for Chinese and Thai languages that aren't supported by nova-3 yet:

- `zh`, `zh-CN`, `zh-Hans` (Mandarin Simplified)
- `zh-TW`, `zh-Hant` (Mandarin Traditional)
- `zh-HK` (Cantonese)
- `th`, `th-TH` (Thai)

Also added unit tests to cover all these cases. Let me know if you'd like any changes!

PR #592, opened 2026-01-07
Force bot leave for captcha challenge

ryusock:
Check for UI forcing a captcha challenge, which is occurring for some zoom bots. Forces a bot leave, so that fallback measure of joining by native SDK can be done. Creates a new bot event subtype so that these events are identifiable.


bots/zoom_web_bot_adapter/zoom_web_ui_methods.py
```diff
@@ -127,6 +127,7 @@ def wait_to_be_admitted_to_meeting(self):
                 time.sleep(1)
                 raise TimeoutException("User has not entered the meeting")
             except TimeoutException as e:
+                self.check_if_captcha_required()
```
ryusock: Correct me if I am wrong but I think we only want to check for this captcha error if we timeout first

noah-duncan (approved):
Looks good thanks!

PR #610, opened 2026-01-21
Add optional filters and indexes for calendar_events API

dvillaume:
Add optional filters and indexes for calendar_events API

New filters:
- event_id: filter by event ID
- start_time_gte: filter events with start_time >= value (ISO 8601)
- end_time_lte: filter events with end_time <= value (ISO 8601)
- ordering: How to order the events. Defaults to `-updated_at`

Performance improvements:
- Add indexes on start_time and end_time in CalendarEvent model
- Migration 0070 to create the indexes

Deprecates the `updated_after` param, it was renamed to `updated_at_gte`

### DIFF
--- a/bots/tests/test_calendars_api_views.py
+++ b/bots/tests/test_calendars_api_views.py
@@ -170,6 +170,23 @@ class CalendarEventListViewTest(TransactionTestCase):
         # The error message uses the new parameter name
         self.assertIn("updated_at_gte", response.json()["error"])
 
+    def test_default_ordering_is_descending_updated_at(self):
+        """Test that results are ordered by -updated_at by default when no ordering param is passed."""
+        # Set distinct updated_at values for the events
+        now = timezone.now()
+        CalendarEvent.objects.filter(pk=self.event_a1.pk).update(updated_at=now - timedelta(hours=2))
+        CalendarEvent.objects.filter(pk=self.event_a2.pk).update(updated_at=now - timedelta(hours=1))
+
+        # Request without ordering parameter
+        response = self._make_authenticated_request("GET", "/api/v1/calendar_events", self.api_key_a_plain)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        results = response.json().get("results", [])
+        event_ids = [e["id"] for e in results]
+
+        # event_a2 was updated more recently, so it should come first (descending order)
+        self.assertEqual(event_ids.index(self.event_a2.object_id) < event_ids.index(self.event_a1.object_id), True)
+
     def test_ordering_parameter(self):
         """Test that ordering parameter correctly orders results."""
         response = self._make_authenticated_request("GET", "/api/v1/calendar_events?ordering=start_time", self.api_key_a_plain)


bots/models.py
```diff
@@ -372,6 +372,8 @@ class CalendarEvent(models.Model):
 
     meeting_url = models.CharField(max_length=511, null=True, blank=True)
 
+    start_time = models.DateTimeField(db_index=True)
+    end_time = models.DateTimeField(db_index=True)
```
noah-duncan, line 388: I think you need to delete the two lines below these, right? Otherwise you are defining the fields twice.

bots/calendars_api_views.py
```diff
@@ -268,7 +268,7 @@ def delete(self, request, object_id):
 
 
 class CalendarEventCursorPagination(CursorPagination):
-    ordering = "-updated_at"
+    ordering = "start_time"
```
noah-duncan: Could you do this in a way so that the ordering is controlled via a query param that defaults to `-updated_at`? Just so that the existing behavior is preserved. I think DRF supports this out of the box: https://www.django-rest-framework.org/api-guide/filtering/#orderingfilter

bots/calendars_api_views.py
```diff
@@ -350,7 +374,38 @@ def get(self, request):
...
+                    return Response({"error": "Invalid start_time_after format. Use ISO 8601 format (e.g., 2025-01-13T10:30:00Z)"}, status=status.HTTP_400_BAD_REQUEST)
+                events = events.filter(start_time__gte=start_time_after_dt)
+            except ValueError:
+                return Response({"error": "Invalid start_time_after format. Use ISO 8601 format (e.g., 2025-01-13T10:30:00Z)"}, status=status.HTTP_400_BAD_REQUEST)
+
+        # Apply end_time_before filter if provided
+        end_time_before = request.query_params.get("end_time_before")
+        if end_time_before is not None:
+            try:
+                from django.utils.dateparse import parse_datetime
```
noah-duncan: Do you need the inline import?

bots/models.py
dvillaume: I removed the older declaration, sorry

bots/calendars_api_views.py
dvillaume: Done! I've added an 'ordering' query parameter that defaults to '-updated_at'.

bots/calendars_api_views.py
dvillaume: I coiped it from existing 'updated_after', but makes sense to move it on top

bots/migrations/0073_alter_calendarevent_end_time_and_more.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 5.1.14 on 2025-01-20 00:00
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
noah-duncan, line 6: Hey @dvillaume I think this migration is funny, maybe because when you generated the file this issue was present: https://github.com/attendee-labs/attendee/pull/610/changes#r2714022437

Can you revert the migration, then delete the file, then recreate the migration? 

When I do this, django produces a slightly different migration
```
# Generated by Django 5.1.2 on 2026-01-23 14:13

from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        ('bots', '0072_backfill_request_sync_for_all_calendars'),
    ]

    operations = [
        migrations.AlterField(
            model_name='calendarevent',
            name='end_time',
            field=models.DateTimeField(db_index=True),
        ),
        migrations.AlterField(
            model_name='calendarevent',
            name='start_time',
            field=models.DateTimeField(db_index=True),
        ),
    ]

```

noah-duncan: @dvillaume I went ahead and updated the migration. You'll want to revert the migration on your local machine and then pull my changes.

dvillaume: Done, thanks Noah.

PR #614, opened 2026-01-22
Upgrade chrome to 144 experiment

noah-duncan:
We needed to lazily load the chrome <> python websocket connection for google meet to work.


rafaeljusto (dismissed):
I can confirm this is working 👍  Nice one.

PR #618, opened 2026-01-23
permit webhook custom url  for private setups

dvillaume:
permit webhook custom url  for private setups

### DIFF
--- a/bots/bots_api_utils.py
+++ b/bots/bots_api_utils.py
@@ -48,7 +48,7 @@ def build_site_url(path=""):
     """
     Build a full URL using SITE_DOMAIN setting.
     Automatically uses http:// for localhost, https:// for everything else.
-    
+
     If EXTERNAL_WEBHOOK_SITE_DOMAIN is set, it takes priority (useful for webhooks that need
     to be accessible from external services like Microsoft/Google).
     """


bots/bots_api_utils.py
```diff
@@ -48,9 +48,14 @@ def build_site_url(path=""):
...
     Build a full URL using SITE_DOMAIN setting.
     Automatically uses http:// for localhost, https:// for everything else.
+    
+    If WEBHOOK_SITE_DOMAIN is set, it takes priority (useful for webhooks that need
+    to be accessible from external services like Microsoft/Google).
     """
-    protocol = "http" if settings.SITE_DOMAIN.startswith("localhost") else "https"
-    return f"{protocol}://{settings.SITE_DOMAIN}{path}"
+    # Use WEBHOOK_SITE_DOMAIN if set (for external webhooks), otherwise use SITE_DOMAIN
+    site_domain = os.getenv("WEBHOOK_SITE_DOMAIN") or settings.SITE_DOMAIN
```
noah-duncan: This looks good, but could you name it `EXTERNAL_WEBHOOK_SITE_DOMAIN`? In the code base we refer to webhooks that go **from** Attendee as 'webhooks' and webhooks that go **to** Attendee from an external server as 'external webhooks'. So want to be consistent with that.

dvillaume: Done, make sense.

PR #630, opened 2026-01-29
Output video allow loop

ryusock:
Support looping of videos with passed in parameter. Some additional logic added in MP4 Demuxer to support looping in the Zoom native SDK, can revert if you prefer.

### DIFF
--- a/bots/zoom_rtms_adapter/zoom_rtms_adapter.py
+++ b/bots/zoom_rtms_adapter/zoom_rtms_adapter.py
@@ -955,6 +955,7 @@ class ZoomRTMSAdapter(BotAdapter):
         return False
 
     def send_video(self, video_url, loop=False):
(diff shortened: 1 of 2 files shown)


bots/zoom_bot_adapter/mp4_demuxer.py
```diff
@@ -28,16 +28,19 @@ class MP4Demuxer:
...
+    def __init__(self, url, output_video_dimensions, on_video_sample, on_audio_sample, loop=False):
         Gst.init(None)
         self._url = url
         self._video_cb = on_video_sample
         self._audio_cb = on_audio_sample
         self._output_video_dimensions = output_video_dimensions
+        self._loop_video = loop
         self._playing = False
-        self._loop = GObject.MainLoop()
+        self._glib_main_loop = GObject.MainLoop()
```
ryusock, line 43: Renaming to avoid confusion between different meanings for "loop" here

PR #660, opened 2026-02-10
Allow PATCH /api/v1/bots/{id} to update bot name and image

dvillaume:
This adds support for updating bot_name and bot_image via PATCH /api/v1/bots/{object_id}.

Clients can change the bot’s display name or avatar after creation, including when bots are already scheduled, so they don’t have to delete and recreate bots when only name or image changes.

- bot_name and bot_image are optional and can be updated in any bot state (same behaviour as metadata).
- bot_image uses the same format as POST /api/v1/bots (type + base64 data).

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -219,9 +219,9 @@ paths:
           description: Bot not found
     patch:
       operationId: Patch Bot
-      description: Updates a bot. join_at, meeting_url, bot_name and bot_image can
-        only be updated when the bot is in the scheduled state. metadata can be updated at any
-        time.
+      description: Updates a bot. The join_at, meeting_url, bot_name and bot_image
+        fields can only be updated when the bot is in the scheduled state. The metadata
+        field can be updated at any time.
       summary: Update a bot
       parameters:
       - in: header


bots/bots_api_utils.py
```diff
@@ -398,6 +398,8 @@ def patch_bot(bot: Bot, data: dict) -> tuple[Bot | None, dict | None]:
         previous_meeting_url = bot.meeting_url
         bot.meeting_url = validated_data.get("meeting_url", bot.meeting_url)
         bot.metadata = validated_data.get("metadata", bot.metadata)
+        if "bot_name" in validated_data and validated_data["bot_name"] is not None:
+            bot.name = validated_data["bot_name"]
 
         # If the join_at or meeting_url is being updated, the state must be scheduled. If it isn't error out.
         update_only_legal_for_scheduled_bots = bot.join_at != previous_join_at or bot.meeting_url != previous_meeting_url
```
noah-duncan: Updating the bot image and bot name should also only be legal in the scheduled state. You can't usually change your name in the middle of the meeting. And if you want to update your image in the middle of the meeting, you should use this endpoint: https://docs.attendee.dev/api-reference/tag/bots/post/api/v1/bots/object_id/output_image

bots/bots_api_utils.py
```diff
@@ -406,6 +408,9 @@ def patch_bot(bot: Bot, data: dict) -> tuple[Bot | None, dict | None]:
 
         bot.save()
 
+        if validated_data.get("bot_image"):
+            create_bot_media_request_for_image(bot, validated_data["bot_image"])
```
noah-duncan: I think you'll want to check if an existing BotMediaRequest is attached to the bot and if so, delete it.

bots/bots_api_utils.py
dvillaume: Thanks Noah, always helpful feedback, much appreciated.

bots/bots_api_utils.py
```diff
@@ -392,6 +392,15 @@ def patch_bot(bot: Bot, data: dict) -> tuple[Bot | None, dict | None]:
     validated_data = serializer.validated_data
 
     try:
+        # bot_name and bot_image can only be updated when the bot is scheduled (use output_image endpoint during meeting)
+        updating_name_or_image = (
+            ("bot_name" in validated_data and validated_data["bot_name"] is not None) or validated_data.get("bot_image")
+        )
+        if updating_name_or_image and bot.state != BotStates.SCHEDULED:
+            return None, {
+                "error": f"Bot is in state {BotStates.state_to_api_code(bot.state)}. bot_name and bot_image can only be updated when the bot is in the scheduled state. To update the bot's image during a meeting, use the POST /api/v1/bots/{{object_id}}/output_image endpoint."
```
noah-duncan: Can this stuff be moved to line 415, where we are doing something similar already?

bots/bots_api_utils.py
```diff
@@ -409,6 +418,8 @@ def patch_bot(bot: Bot, data: dict) -> tuple[Bot | None, dict | None]:
         bot.save()
 
         if validated_data.get("bot_image"):
+            # Replace any existing image request for this bot (scheduled bots have at most one pending image)
+            BotMediaRequest.objects.filter(bot=bot, media_type=BotMediaRequestMediaTypes.IMAGE).delete()
```
noah-duncan: this is the same thing, I just like how it looks better and good to make it clear we're only deleting enqueued ones. `bot.media_requests.filter(media_type=BotMediaRequestMediaTypes.IMAGE, state=BotMediaRequestStates.ENQUEUED).delete()`

PR #708, opened 2026-03-04
feat: Add Sentry integration for error tracking and performance monitoring

illing2005:
## Summary

- Adds optional Sentry integration for error tracking, performance monitoring, and Celery task monitoring
- Controlled entirely via environment variables - disabled when `SENTRY_DSN` is not set
- Includes Django, Celery, and Redis integrations

## Environment Variables

| Variable | Description | Default |
|----------|-------------|---------|
| `SENTRY_DSN` | Sentry project DSN (required to enable) | None |
| `SENTRY_TRACES_SAMPLE_RATE` | Performance sampling rate (0.0-1.0) | 0.1 |
| `SENTRY_PROFILES_SAMPLE_RATE` | Profiling sampling rate (0.0-1.0) | 0.1 |
| `SENTRY_ENVIRONMENT` | Environment name | production |
| `SENTRY_SEND_PII` | Send personally identifiable info | false |

## Test plan

- [x] Verify app starts without `SENTRY_DSN` set (Sentry disabled)
- [ ] Verify app starts with `SENTRY_DSN` set (Sentry enabled)
- [ ] Verify Celery workers initialize Sentry correctly


attendee/celery.py
```diff
@@ -4,9 +4,14 @@
 
 from celery import Celery
 
+from attendee.sentry import init_sentry
+
 # Set the default Django settings module
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "attendee.settings")
 
+# Initialize Sentry for Celery workers (only if SENTRY_DSN is set)
+init_sentry()
```
noah-duncan: Do you need this in both base.py and celery.py?

illing2005: Yes - Celery workers run as separate processes and don't always load Django settings. The init is idempotent so duplicate calls are safe.

noah-duncan: Ok, in our case the celery config is being driven by django settings, so I don't think you need it. See this line: https://github.com/attendee-labs/attendee/blob/a98e439dea2c35d96024999a9c5ca1126061c5d1/attendee/celery.py#L39

illing2005: Good point, removed it.

noah-duncan (approved):
Thanks @illing2005 , looks good!

PR #718, opened 2026-03-11
feat: Zoom webinar support

ryusock:
Add ability for Attendee bots to join Zoom webinars.

Requires webinar license to create webinars, and then to promote bots in webinars to **panelist** (see [webinar role definitions and abilities](https://support.zoom.com/hc/en/article?id=zm_kb&sysparm_article=KB0065551)). Note that only bots that have been promoted from **attendee** to **panelist** can record meetings, either in the webinar by the host or by inviting the bot as panelist (must be a signed-in bot). Bots, or more precisely the accounts that the bot is associated with, may also have recording permission if they are added as co-host, but co-hosts can only be added from within the same Zoom business organization. This is a Zoom limitation.

Supports 4 different webinar join path combinations, 1) with or without registration, and 2) anonymous or signed-in bot.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,3 +99,4 @@ db.sqlite3
 
 custom/
 docker-compose.yaml
+.idea/
--- a/bots/app_session_api_utils.py
+++ b/bots/app_session_api_utils.py
@@ -7,7 +7,7 @@ from django.db import IntegrityError, transaction
 from .app_session_serializers import (
     CreateAppSessionSerializer,
 )
-from .bots_api_utils import BotCreationSource, create_webhook_subscriptions
+from .bots_api_utils import BotCreationSource, create_webhook_subscriptions, validate_bot_concurrency_limit, validate_external_media_storage_settings
 from .models import (
     Bot,
     BotEventManager,
@@ -37,6 +37,7 @@ def create_app_session(data: dict, source: BotCreationSource, project: Project) 
     rtmp_settings = serializer.validated_data["rtmp_settings"]
     recording_settings = serializer.validated_data["recording_settings"]
     debug_settings = serializer.validated_data["debug_settings"]
+    external_media_storage_settings = serializer.validated_data["external_media_storage_settings"]
 
     metadata = serializer.validated_data["metadata"]
     websocket_settings = serializer.validated_data["websocket_settings"]
@@ -45,12 +46,21 @@ def create_app_session(data: dict, source: BotCreationSource, project: Project) 
     zoom_rtms = serializer.validated_data["zoom_rtms"]
     initial_state = BotStates.READY
 
+    error = validate_external_media_storage_settings(external_media_storage_settings, project)
+    if error:
+        return None, error
+
+    error = validate_bot_concurrency_limit(project)
+    if error:
+        return None, error
+
     settings = {
         "transcription_settings": transcription_settings,
         "rtmp_settings": rtmp_settings,
         "recording_settings": recording_settings,
         "debug_settings": debug_settings,
         "websocket_settings": websocket_settings,
+        "external_media_storage_settings": external_media_storage_settings,
         "zoom_rtms": zoom_rtms,
     }
 
--- a/bots/redis_utils.py
+++ b/bots/redis_utils.py
@@ -1,0 +1,30 @@
+# Lua fallback for Redis < 7 which doesn't support EXPIRE ... NX.
+_redis_lua_script_incr_and_expire_nx = None
+
+
+def _get_redis_lua_script_incr_and_expire_nx(redis_client):
+    """Lua fallback for Redis < 7 which doesn't support EXPIRE ... NX."""
+    global _redis_lua_script_incr_and_expire_nx
+    if _redis_lua_script_incr_and_expire_nx is None:
+        _redis_lua_script_incr_and_expire_nx = redis_client.register_script(
+            """
+            -- incr_and_expire_nx: INCR key, set EXPIRE only on first creation
+            local count = redis.call('INCR', KEYS[1])
+            local ttl_set = 0
+            if count == 1 then
+                ttl_set = redis.call('EXPIRE', KEYS[1], ARGV[1])
+            end
+            return {count, ttl_set}
+            """
+        )
+    return _redis_lua_script_incr_and_expire_nx
+
+
+def incr_and_expire_nx(redis_client, key, ttl):
+    """Atomically INCR a key and set its TTL only if the key is new.
+
+    Returns (count, ttl_set) where ttl_set indicates whether EXPIRE was applied.
+    """
+    script = _get_redis_lua_script_incr_and_expire_nx(redis_client)
+    count, ttl_set = script(keys=[key], args=[ttl])
+    return count, ttl_set
--- a/bots/tasks/deliver_webhook_task.py
+++ b/bots/tasks/deliver_webhook_task.py
@@ -9,6 +9,7 @@ from django.conf import settings
 from django.utils import timezone
 
 from bots.models import SessionTypes, WebhookDeliveryAttempt, WebhookDeliveryAttemptStatus, WebhookTriggerTypes
+from bots.redis_utils import incr_and_expire_nx
 from bots.webhook_utils import sign_payload
 
 logger = logging.getLogger(__name__)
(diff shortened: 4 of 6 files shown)


bots/zoom_bot_adapter/zoom_bot_adapter.py
```diff
@@ -96,7 +96,7 @@ def __init__(
         self.record_participant_speech_start_stop_events = record_participant_speech_start_stop_events
 
         self._jwt_token = generate_jwt(zoom_client_id, zoom_client_secret)
-        self.meeting_id, self.meeting_password = parse_zoom_join_url(meeting_url)
+        self.meeting_id, self.meeting_password = parse_zoom_join_url(meeting_url)[:2]
```
noah-duncan: The [:2] is awkward. You could either have the `parse_zoom_join_url` method return a dict or just not modify it at all and create a separate dedicated method for extracting the tk param from the url. Latter option sounds best, looks like there's only one place in the code where we need to extract the tk param from the url.

bots/zoom_bot_adapter/zoom_bot_adapter.py
```diff
@@ -648,9 +656,16 @@ def on_local_recording_privilege_request_status_changed(status):
...
                     self.handle_recording_permission_denied(reason=BotAdapter.BOT_RECORDING_PERMISSION_DENIED_REASON.HOST_CLIENT_CANNOT_GRANT_PERMISSION)
-
-                self.recording_ctrl.RequestLocalRecordingPrivilege()
-                logger.info("Requesting recording privilege.")
+                elif is_support_request_local_recording_privilege_result == zoom.SDKERR_WRONG_USAGE:
+                    # Hacky, but we're assuming SDKERR_WRONG_USAGE means we are in a webinar where requesting
+                    # local recording privilege is not applicable (bot joined as attendee, not yet promoted to panelist).
+                    logger.info("Webinar context detected (SDKERR_WRONG_USAGE). Treating as permission denied.")
+                    logger.info("Bot joined as webinar attendee; waiting for host to promote to panelist.")
+                    self.is_webinar = True
```
noah-duncan: I bet there's other reasons why it might return `zoom.SDKERR_WRONG_USAGE` besides being in a webinar. Also, there's a dedicated way to find this info see here: https://chatgpt.com/share/69b8b2ad-8b48-800c-bf04-895f022c9711

bots/zoom_bot_adapter/zoom_bot_adapter.py
```diff
@@ -1030,6 +1049,9 @@ def meeting_status_changed(self, status, iResult):
             self.is_joining_or_leaving_breakout_room = True
             self.send_message_callback({"message": self.Messages.LEAVING_BREAKOUT_ROOM})
 
+        if status == zoom.MEETING_STATUS_RECONNECTING:
+            self.is_reconnecting = True
```
noah-duncan: Do we need this? IME it doesn't stay in the reconnecting state for a significant amount of time

bots/zoom_bot_adapter/zoom_bot_adapter.py
```diff
@@ -943,6 +958,10 @@ def join_meeting(self):
             param.app_privilege_token = self.zoom_tokens.get("app_privilege_token")
         if self.zoom_tokens.get("onbehalf_token"):
             param.onBehalfToken = self.zoom_tokens.get("onbehalf_token")
+        # Set the webinarToken only if joining a webinar as an attendee (in webinars, all attendees are in Guest Mode).
+        # If joining as a signed-in bot (for panelists and co-hosts), use the ZAK token instead, and leave the webinarToken as NULL.
+        if self.zoom_tokens.get("registrant_token") and not self.zoom_tokens.get("zak_token"):
+            param.webinarToken = self.zoom_tokens.get("registrant_token")
```
noah-duncan, line 963: I would add some kind of test that giving the meeting url a tk token will cause this to get set as expected.

bots/zoom_bot_adapter/zoom_bot_adapter.py
ryusock: I think after I change the `is_webinar` to be detected via the proper signal you suggested, I won't need this anymore. I added it because the current method had a gap in detection

bots/zoom_bot_adapter/zoom_bot_adapter.py
```diff
@@ -648,9 +658,12 @@ def on_local_recording_privilege_request_status_changed(status):
...
                 if is_support_request_local_recording_privilege_result == zoom.SDKERR_MEETING_DONT_SUPPORT_FEATURE:
                     self.handle_recording_permission_denied(reason=BotAdapter.BOT_RECORDING_PERMISSION_DENIED_REASON.HOST_CLIENT_CANNOT_GRANT_PERMISSION)
-
-                self.recording_ctrl.RequestLocalRecordingPrivilege()
-                logger.info("Requesting recording privilege.")
+                elif self.is_webinar and is_support_request_local_recording_privilege_result == zoom.SDKERR_WRONG_USAGE:
+                    # SDKERR_WRONG_USAGE from IsSupportRequestLocalRecordingPrivilege in a webinar means the bot is an attendee and needs to be promoted to panelist
+                    self.handle_recording_permission_denied(reason=BotAdapter.BOT_RECORDING_PERMISSION_DENIED_REASON.WEBINAR_ATTENDEE_NEEDS_PANELIST_PROMOTION)
+                else:
+                    self.recording_ctrl.RequestLocalRecordingPrivilege()
```
noah-duncan: Can you keep the original logic where we call `self.recording_ctrl.RequestLocalRecordingPrivilege()` regardless, unless there's a good reason not to?

bots/zoom_bot_adapter/zoom_bot_adapter.py
```diff
@@ -1041,7 +1058,10 @@ def meeting_status_changed(self, status, iResult):
             GLib.timeout_add_seconds(self.automatic_leave_configuration.waiting_room_timeout_seconds, self.leave_meeting_if_still_in_waiting_room)
 
         if status == zoom.MEETING_STATUS_INMEETING:
-            self.send_message_callback({"message": self.Messages.BOT_JOINED_MEETING})
+            if not self.joined_at:
```
noah-duncan: why did we change the logic here?

bots/zoom_bot_adapter/zoom_bot_adapter.py
```diff
@@ -1041,7 +1058,10 @@ def meeting_status_changed(self, status, iResult):
...
+                self.send_message_callback({"message": self.Messages.BOT_JOINED_MEETING})
+            elif self.is_webinar:
+                self.send_message_callback({"message": self.Messages.WEBINAR_BOT_PROMOTED_TO_PANELIST})
```
noah-duncan: This does not look like the most direct way to detect whether someone is promoted to panelist.

bots/zoom_bot_adapter/zoom_bot_adapter.py
ryusock: The reasoning was if joined_at was already set, when the bot rejoined as a panelist there was a warning that displayed in the console. It is because `BOT_JOINED_MEETING` fires again during the rejoin, when it had already fired at the initial join. But this is not a catastrophic error, only a warning, and the bot continues as expected. Here is the error
```
[2026-03-22 18:33:24,079: WARNING/ForkPoolWorker-16] Traceback (most recent call last):
attendee-worker-local-1     |   File "/attendee/bots/bot_controller/bot_controller.py", line 1587, in <lambda>
attendee-worker-local-1     |     GLib.idle_add(lambda: self.take_action_based_on_message_from_adapter(message))
attendee-worker-local-1     |   File "/attendee/bots/bot_controller/bot_controller.py", line 1945, in take_action_based_on_message_from_adapter
attendee-worker-local-1     |     BotEventManager.create_event(bot=self.bot_in_db, event_type=BotEventTypes.BOT_JOINED_MEETING)
attendee-worker-local-1     |   File "/attendee/bots/models.py", line 1766, in create_event
attendee-worker-local-1     |     raise ValidationError(f"Event {BotEventTypes.type_to_api_code(event_type)} not allowed when bot is in state {BotStates.state_to_api_code(old_state)}. It is only allowed in these states: {', '.join(valid_states_labels)}")
attendee-worker-local-1     | django.core.exceptions.ValidationError: ['Event joined_meeting not allowed when bot is in state joined_recording_permission_denied. It is only allowed in these states: waiting_room, joining']
```

But this is really not something worth changing core logic for. So I am changing this back

bots/zoom_bot_adapter/zoom_bot_adapter.py
ryusock: I've decided to use a new flag as a more direct way to detect promotion to panelist. It doesn't replace the existing flow for sending the `BOT_JOINED_MEETING`, only sends an additional promotion message (via bot log entry) if applicable, based on the flag `joined_as_webinar_attendee`

noah-duncan: Can you see if there's a way to directly tell if the user was promoted to panelist instead of doing this?

PR #723, opened 2026-03-12
feat: Realtime per participant audio

noah-duncan:


bots/bot_controller/bot_controller.py
```diff
@@ -119,6 +119,42 @@ def disable_incoming_video_for_web_bots(self):
...
+            return self.send_per_participant_audio_chunk_to_websocket_client
+
+        return None
+
+    def send_per_participant_audio_chunk_to_websocket_client(self, speaker_id, chunk_time, chunk_bytes):
+        if not self.websocket_client_manager:
+            return
+
+        payload = per_participant_audio_websocket_payload(
+            participant_uuid=speaker_id,
```
echaoeoen, line 152: Hi @noah-duncan ,

Would it be possible to include a participant_name field in the audio WebSocket payload?

In our use case, we're building a voice agent assistant that interacts with multiple participants in a session. Having access to the participant name directly in the payload would allow the agent to better identify speakers and produce more natural, context-aware interactions.

Currently, we can infer speakers from transcription or external mapping, but including this metadata at the transport layer would significantly simplify the pipeline and improve real-time responsiveness.

Happy to contribute or help with implementation if this aligns with the project direction.

Thanks!

PR #764, opened 2026-04-01
feat: Refactor signed in teams and gmeet bots to use single model

ryusock:
Unifies Google Meet and Teams signed-in bot credentials under shared BotLoginGroup / BotLogin models, adds multi-group management in the dashboard, and lets bot creation target a specific login group via platform settings. 

Bot login selection is now platform-aware and group-aware. If a bot is created with google_meet_settings.login_group_name or teams_settings.login_group_name, runtime only considers that named group; otherwise it scans groups for that platform in creation order. Within the chosen group, it selects the least recently used active login, so usage rotates over time instead of always reusing the same account.

Also includes a data-model migration that renames the old Google Meet login tables into generic bot login tables, adds platform-aware group uniqueness, introduces object IDs for groups/logins, and backfills Teams bot logins from legacy Credentials rows into the new shared storage.


bots/models.py
```diff
@@ -77,33 +85,54 @@ def save(self, *args, **kwargs):
     def __str__(self):
         return f"{self.project.name} - {self.object_id}"
 
+    @classmethod
+    def first_available_login(cls, project, platform, group_name=None):
+        """
+        Returns the least recently used BotLogin for the specified project and platform.
+
+        - If a group_name is provided, only considers logins in that named group.
+        - If no group_name is given, iterates through all groups (ordered by creation date and id)
```
noah-duncan: considering logins across all groups when none is specified, feels weird, because it's a different mode of operation. Maybe if no group is specified we use the oldest one?

bots/models.py
```diff
@@ -77,33 +85,54 @@ def save(self, *args, **kwargs):
...
+
+        If no valid login is found, returns None.
+        """
+        groups = cls.objects.filter(project=project, platform=platform)
+        if group_name is not None:
+            groups = groups.filter(name=group_name)
+        groups = groups.order_by("created_at", "id")
+
+        for group in groups:
+            available_login = group.bot_logins.order_by(F("last_used_at").asc(nulls_first=True), "id").first()
```
noah-duncan: ok, so it tiebreaks on id, nice

bots/models.py
```diff
@@ -130,9 +159,10 @@ def __str__(self):
         return f"{self.email} - {self.object_id}"
 
     class Meta:
-        # Within a Google Meet Bot Login Group, we don't want to allow Google Meet Bot Logins with the same email
+        db_table = "bots_googlemeetbotlogin"
+        # Within a BotLoginGroup, we don't want to allow BotLogins with the same email
         constraints = [
-            models.UniqueConstraint(fields=["group", "email"], name="unique_google_meet_bot_login_email"),
+            models.UniqueConstraint(fields=["group", "email"], name="unique_bot_login_email_per_group"),
```
noah-duncan: Can you research if there could be any issue with bots running the old code from renaming the constraint? If there is any risk at all, then would prefer not to just like we kept the table name the same. It doesn't really impact anything, we can just leave a comment that we kept the name to avoid any db issues during the backfill.

bots/models.py
```diff
@@ -77,33 +85,54 @@ def save(self, *args, **kwargs):
...
+            if available_login:
+                return available_login
+
+        return None
 
-class GoogleMeetBotLogin(models.Model):
-    OBJECT_ID_PREFIX = "gbl_"
-    group = models.ForeignKey(GoogleMeetBotLoginGroup, on_delete=models.CASCADE, related_name="google_meet_bot_logins")
+    class Meta:
+        db_table = "bots_googlemeetbotlogingroup"
```
noah-duncan: Would be good to add a comment explaining this

bots/projects_views.py
```diff
@@ -143,13 +173,12 @@ def get_partial_for_credential_type(credential_type, request, context):
...
     elif credential_type == Credentials.CredentialTypes.ELEVENLABS:
         return render(request, "projects/partials/elevenlabs_credentials.html", context)
-    elif credential_type == Credentials.CredentialTypes.TEAMS_BOT_LOGIN:
-        return render(request, "projects/partials/teams_bot_login_credentials.html", context)
     elif credential_type == Credentials.CredentialTypes.KYUTAI:
         return render(request, "projects/partials/kyutai_credentials.html", context)
     elif credential_type == Credentials.CredentialTypes.EXTERNAL_MEDIA_STORAGE:
         return render(request, "projects/partials/external_media_storage_credentials.html", context)
     else:
+        # Credential type TEAMS_BOT_LOGIN is no longer supported in CreateCredentialsView
```
noah-duncan: idk if we need this left in the codebase

bots/projects_views.py
```diff
@@ -290,7 +319,7 @@ def post(self, request, object_id):
 
         try:
             credential_type = int(request.POST.get("credential_type"))
-            if credential_type not in [choice[0] for choice in Credentials.CredentialTypes.choices]:
+            if credential_type not in [choice[0] for choice in Credentials.CredentialTypes.choices] or credential_type == Credentials.CredentialTypes.TEAMS_BOT_LOGIN:
```
noah-duncan: I think its ok to just let this slide...maybe will be confusing for one person, but it makes the code look funny forever. Or you can set a reminder to clean it up a month later or something

bots/projects_views.py
```diff
@@ -386,7 +410,7 @@ def post(self, request, object_id):
...
```
noah-duncan: dont think we need this kind of thing at all for the deletion case

bots/projects_views.py
```diff
@@ -1409,13 +1425,96 @@ def patch(self, request, object_id):
...
 class CreateGoogleMeetBotLoginView(LoginRequiredMixin, ProjectUrlContextMixin, View):
     def post(self, request, object_id):
         project = get_project_for_user(user=request.user, project_object_id=object_id)
+        bot_login_group_object_id = request.POST.get("bot_login_group_object_id")
+        google_meet_bot_login_group = get_bot_login_group_for_user(user=request.user, bot_login_group_object_id=bot_login_group_object_id)
 
         try:
-            # Get or create GoogleMeetBotLoginGroup for this project
-            google_meet_bot_login_group, created = GoogleMeetBotLoginGroup.objects.get_or_create(project=project)
+            if google_meet_bot_login_group.project_id != project.id or google_meet_bot_login_group.platform != BotLoginPlatform.GOOGLE_MEET:
```
noah-duncan: Maybe its better to pass the platform and project into `get_bot_login_group_for_user`? Then you don't need these checks, right?

bots/serializers.py
```diff
@@ -714,6 +714,11 @@ class MetadataJSONField(serializers.JSONField):
             "description": "The mode to use for the Google Meet bot login. 'always' means the bot will always login, 'only_if_required' means the bot will only login if the meeting requires authentication.",
             "default": "always",
         },
+        "login_group_name": {
+            "type": ["string", "null"],
+            "description": "Optional bot login group name to use for Google Meet signed-in bot selection.",
```
noah-duncan: Should be mentioned somewhere that if no group is specified, the oldest one will be used.

bots/migrations/0081_backfill_botlogingroup_and_botlogin_and_more.py
```diff
@@ -0,0 +1,82 @@
...
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+def backfill_bot_login_groups(apps, schema_editor):
+    BotLoginGroup = apps.get_model("bots", "BotLoginGroup")
+    BotLogin = apps.get_model("bots", "BotLogin")
+
+    for project_id in BotLoginGroup.objects.values_list("project_id", flat=True).distinct():
+        groups = BotLoginGroup.objects.filter(project_id=project_id).order_by("id")
```
noah-duncan: Do we need this backfill? 

Don't think we need to set the name, because the old UI only allowed one login group to be created.

I think its ok if there are old style object_id's in the database.

If you have the platform default to google meet then you don't need to worry about setting that either.

bots/migrations/0082_backfill_teams_bot_logins.py
```diff
@@ -0,0 +1,63 @@
...
+        username = credentials_dict.get("username")
+        password = credentials_dict.get("password")
+
+        if not username or not password:
+            continue
+
+        bot_login_group, created = BotLoginGroup.objects.get_or_create(
+            project=credential.project,
+            platform="teams",
+            name="Teams Group 1",
```
noah-duncan: Do we need to set the name?

bots/teams_bot_adapter/teams_ui_methods.py
```diff
@@ -268,7 +268,11 @@ def get_layout_to_select(self):
 
     # Returns nothing if succeeded, raises an exception if failed
     def attempt_to_join_meeting(self):
-        if self.teams_bot_login_credentials and self.teams_bot_login_should_be_used:
+        # REVIEW: This is following the same pattern as the Google Meet bot adapter, but it appears incorrect.
```
noah-duncan: Oh yea it's a little confusing. I think if you look at the google code you'll see that `google_bot_login_should_be_used` will only ever be true if there is a login available.

bots/teams_bot_adapter/teams_ui_methods.py
```diff
@@ -371,13 +375,18 @@ def click_cancel_join_button(self):
         time.sleep(1)
 
     def login_to_microsoft_account(self):
+        credentials = self.create_teams_bot_login_credentials_callback()
```
noah-duncan: Should be called fetch credentials instead of create credentials probably. It just works a bit differently than for google

bots/teams_bot_adapter/teams_bot_adapter.py
```diff
@@ -49,20 +50,22 @@ def __init__(
...
+        create_teams_bot_login_credentials_callback: Callable[[], dict],
         modify_dom_for_video_recording: bool,
         **kwargs,
     ):
         super().__init__(*args, **kwargs)
         self.teams_closed_captions_language = teams_closed_captions_language
-        self.teams_bot_login_credentials = teams_bot_login_credentials
-        self.teams_bot_login_should_be_used = teams_bot_login_should_be_used and teams_bot_login_credentials
+        self.teams_bot_login_is_available = teams_bot_login_is_available
+        self.teams_bot_login_should_be_used = teams_bot_login_should_be_used
```
noah-duncan: This should follow the pattern we use for the google bot to ensure that "should_be_used" is false if the credentials aren't available `        self.google_meet_bot_login_should_be_used = google_meet_bot_login_should_be_used and google_meet_bot_login_is_available`

bots/migrations/0082_backfill_teams_bot_logins.py
ryusock: Yes, we have a requirement that the name is not null, so need to set on here. I can make it optional, but then if there is an existing no name group for that platform, we won't be able to create a new group unless it is named.

Or are you asking if there should be a default name for this backfill specifically?

noah-duncan: Ok, forgot it was non null. That's fine then.

bots/migrations/0081_backfill_botlogingroup_and_botlogin_and_more.py
noah-duncan: Ok, I guess we need the backfill because we need to set the name. I wouldn't bother redoing the object_ids though.

ryusock: Yep, keeping the backfill without the object id changes

bots/models.py
ryusock: Reverted the uniqueness constraint rename to avoid potential volatility

bots/templates/projects/project_credentials.html
```diff
@@ -21,9 +21,12 @@ <h2>Credentials</h2>
...
     {% include 'projects/partials/kyutai_credentials.html' with credential_type=kyutai_credential_type credentials=kyutai_credentials %}
     {% include 'projects/partials/google_tts_credentials.html' with credential_type=google_tts_credential_type credentials=google_tts_credentials %}
-    {% include 'projects/partials/teams_bot_login_credentials.html' with credential_type=teams_bot_login_credential_type credentials=teams_bot_login_credentials %}
-    {% include 'projects/partials/google_meet_bot_login_group.html' with google_meet_bot_login_group=google_meet_bot_login_group %}
     {% include 'projects/partials/external_media_storage_credentials.html' with credential_type=external_media_storage_credential_type credentials=external_media_storage_credentials %}
+    <div class="alert alert-info mt-4 mb-0" role="alert">
+        <i class="bi bi-info-circle me-2"></i>
+        Google Meet and Teams account credentials for signed-in bots are now managed on the
+        <a href="{% url 'projects:project-bot-logins' project.object_id %}" class="alert-link">Bot Logins page</a>.
```
noah-duncan, line 28: Thanks, this is thoughtful

bots/models.py
```diff
@@ -2573,7 +2609,6 @@ class CredentialTypes(models.IntegerChoices):
         OPENAI = 5, "OpenAI"
         ASSEMBLY_AI = 6, "Assembly AI"
         SARVAM = 7, "Sarvam"
-        TEAMS_BOT_LOGIN = 8, "Teams Bot Login"
```
noah-duncan, line 2576: Don't think we want to delete this even though it isn't used anymore. Could just confusion if we somehow get an old database, it's nice to know what the value '8' maps to

ryusock: Got it

noah-duncan (dismissed):
Thanks @ryusock

PR #783, opened 2026-04-10
Security: Wildcard ALLOWED_HOSTS in GKE production/staging settings

tomaioo:
## Summary

Security: Wildcard ALLOWED_HOSTS in GKE production/staging settings

## Problem

**Severity**: `Medium` | **File**: `attendee/settings/production-gke.py:L10`

Both GKE deployment settings use `ALLOWED_HOSTS = ["*"]`, disabling Django's host header protection in those environments as well. Even behind ingress, defense-in-depth recommends validating hosts at the app layer.

## Solution

Use explicit hostnames per environment (e.g., `api.attendee.dev`, `staging.attendee.dev`) loaded from environment variables. Mirror this change in `attendee/settings/staging-gke.py`.

## Changes

- `attendee/settings/production-gke.py` (modified)

### DIFF
--- a/attendee/settings/production-gke.py
+++ b/attendee/settings/production-gke.py
@@ -7,7 +7,7 @@ from .base import *
 from .base import LOG_FORMATTERS
 
 DEBUG = False
-ALLOWED_HOSTS = ["*"]
+ALLOWED_HOSTS = os.getenv("ALLOWED_HOSTS", "api.attendee.dev").split(",")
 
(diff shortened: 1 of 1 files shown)


noah-duncan:
Thanks for the PR @tomaioo! Your PR is not backwards compatible, so I decided to implement it slightly differently here: https://github.com/attendee-labs/attendee/pull/785

The concern is valid though.

PR #786, opened 2026-04-13
feat: Add mute video param to output video endpoint

ryusock:


bots/zoom_bot_adapter/zoom_bot_adapter.py
```diff
@@ -1237,7 +1237,7 @@ def send_video(self, video_url, loop=False):
             url=video_url,
             output_video_dimensions=(self.suggested_video_cap.width, self.suggested_video_cap.height),
             on_video_sample=self.mp4_demuxer_on_video_sample,
-            on_audio_sample=self.mp4_demuxer_on_audio_sample,
+            on_audio_sample=None if mute_video else self.mp4_demuxer_on_audio_sample,
```
noah-duncan: It doesn't cause any problems if you pass `on_audio_sample=None` into the `MP4Demuxer` class? Reading the code, kind of looks like it would crash

ryusock: It didn't in my testing, but I'll verify that to be certain

PR #801, opened 2026-04-28
isolate web bot recording audio

joaotolovi:
This fixes mixed audio in final meeting recordings when multiple web bots are running at the same time.

Root cause: the final recorder captured audio from the default ALSA/PulseAudio device. When simultaneous meetings shared the same PulseAudio sink/source, the final recording could include audio from other meetings even though the video was correct.

Changes:

Create a dedicated PulseAudio null sink per web bot recording.
Route Chrome audio for that bot to its dedicated sink via PULSE_SINK.
Make ffmpeg record from that sink’s monitor instead of the global default device.
Mute/unmute only the bot-specific sink during pause/resume.
Unload the dedicated sink during cleanup.
Add unit coverage for sink setup, ffmpeg input selection, fallback behavior, pause muting, and cleanup.

### DIFF
--- a/bots/bot_controller/bot_controller.py
+++ b/bots/bot_controller/bot_controller.py
@@ -186,6 +186,8 @@ class BotController:
     def get_google_meet_bot_adapter(self):
         from bots.google_meet_bot_adapter import GoogleMeetBotAdapter
 
+        browser_audio_environment = self.screen_and_audio_recorder.browser_audio_environment() if self.screen_and_audio_recorder else {}
+
         return GoogleMeetBotAdapter(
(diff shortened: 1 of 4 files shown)


noah-duncan:
@joaotolovi thanks, this is smart. Running bots on celery has other issues that make it unsuitable for use in production, such as that when you deploy or rescale, running bots will get terminated when the celery containers are shut down. So using k8s is still the better option, so merging this is a fairly low priority as it does add some complexity to the codebase.

But will refer people who strongly prefer to use celery to this PR.

PR #833, opened 2026-05-22
feat: add delete_old_bot_transcripts management command

rdoria1:
Bulk-cleanup variant of the existing Delete Bot Transcript endpoint. Deletes Utterance rows for bots created more than --days days ago (default 14). Other tables (Bot, Recording, AudioChunk, Participant, etc.) are left intact,

Includes --dry-run for safe verification and --batch-size to keep transactions short and avoid long-held locks on bots_utterance.

Mirrors the existing pattern in delete_old_audio_chunks.


bots/management/commands/delete_old_bot_transcripts.py
```diff
@@ -0,0 +1,86 @@
...
+        dry_run = options["dry_run"]
+        batch_size = options["batch_size"]
+
+        cutoff = timezone.now() - timedelta(days=days)
+        logger.info(
+            f"Looking for bots created before {cutoff.isoformat()} ({days} days ago)..."
+        )
+
+        old_bot_ids = list(
+            Bot.objects.filter(created_at__lt=cutoff).values_list("id", flat=True)
```
noah-duncan: I wouldn't filter on created_at because when using scheduled bots, a bot can be created a long time before it actually runs.

Instead, can you filter on when the bot ended? We unfortunately don't have a column for this on the bots table, but you can filter on the bot events table and still get the bot id from that.

```
BotEvent.objects.filter(new_state=BotStates.ENDED, created_at__lt=cutoff).values_list("bot_id", flat=True)
```

noah-duncan (approved):

noah-duncan (changes requested):
Thanks @rdoria1 looks great, please just run `ruff check --fix; ruff format` to lint the PR so the tests pass

noah-duncan (approved):

PR #842, opened 2026-05-29
feat: add INTERNAL_SITE_DOMAIN for in-cluster bot callbacks

rafaeljusto:
build_site_url is used for two distinct audiences: URLs that must be reachable from outside the cluster (webhook addresses registered with external services, dashboard links shown to users) and callbacks that bot pods make back to the app from inside the cluster (e.g. the Google Meet set-cookie URL the bot's browser navigates to).

When bot pods are subject to a restrictive NetworkPolicy that blocks the public ingress, only the internal callback needs to resolve to an in-cluster address; the external URLs must stay public.

Add a dedicated build_internal_site_url function and a new optional INTERNAL_SITE_DOMAIN env var that internal callbacks resolve against over http:// (e.g. service.namespace.svc.cluster.local:8000), so an operator can target an in-cluster service and bypass the ingress. Route the Google Meet set-cookie URL through build_internal_site_url.

Backward compatible: when INTERNAL_SITE_DOMAIN is unset, build_internal_site_url falls back to build_site_url, preserving the existing EXTERNAL_WEBHOOK_SITE_DOMAIN / SITE_DOMAIN behavior, so current deployments are unaffected.


bots/bots_api_utils.py
```diff
@@ -45,13 +45,35 @@
...
+        http://attendee-app.ai.svc.cluster.local:8000) so the operator can target an
+        in-cluster service and bypass the public ingress, which is useful when bot pods are
+        subject to a restrictive NetworkPolicy. When unset, internal callbacks fall back to
+        the external domain, preserving the previous behaviour.
     """
+    if internal:
+        bot_callback_domain = os.getenv("BOT_CALLBACK_SITE_DOMAIN")
+        if bot_callback_domain:
+            # Allow an optional protocol prefix so the operator can pick http for in-cluster services
+            if bot_callback_domain.startswith(("http://", "https://")):
```
noah-duncan: Hey @rafaeljusto overall this looks fine. But is there ever a case when the local callback URL will use https? Can we simplify the logic and just add http every time?

rafaeljusto: Yes we can! I was trying to be flexible as I may be missing some use-case 😄  I will patch it there. 👍

rafaeljusto: Done!

noah-duncan (approved):
Thanks @rafaeljusto ! Made a few changes. The env var was renamed to `INTERNAL_SITE_DOMAIN`, since `BOT_CALLBACK_SITE_DOMAIN` was referring to the specific use-case it is used for and we may want to use this for other use cases in the future.

PR #844, opened 2026-06-01
fix: support DISABLE_SERVER_SIDE_CURSORS env var for pgBouncer

pylead:
Support `DISABLE_SERVER_SIDE_CURSORS` env var to avoid login errors on postgres with `pool_mode=transaction`
(e.g. PgBouncer in `pool_mode=transaction`, Cloud SQL Auth Proxy in tx mode, etc.)

Currently, on login Django's server-side cursors fail intermittently:

```
psycopg2.errors.InvalidCursorName: cursor "_django_curs_..." does not exist
```

This is documented behavior — Django's docs recommend `DISABLE_SERVER_SIDE_CURSORS = True` for anyone running behind a transaction-pooling pooler: https://docs.djangoproject.com/en/4.2/ref/databases/#transaction-pooling-server-side-cursors


noah-duncan (approved):
Thanks @pylead !

PR #846, opened 2026-06-02
feat(sarvam): add translation support and newer model

RohanDebroy:
**Summary**
- Added support for the new model: `saaras:v3`
- Added a new `mode` parameter for Saram, enabling translation functionality

**Testing**
- Used AI assistance for:
  - Writing the test cases
  - Understanding repository context and codebase knowledge
- Manually tested all relevant flows to verify functionality and behavior

Screenshot 2026-06-02 at 3 18 13 PM


bots/serializers.py
```diff
@@ -1168,6 +1173,10 @@ def validate_transcription_settings(self, value):
         if ("custom_async" in value or "custom_async_v2" in value) and not os.getenv("CUSTOM_ASYNC_TRANSCRIPTION_URL"):
             raise serializers.ValidationError({"transcription_settings": "CUSTOM_ASYNC_TRANSCRIPTION_URL environment variable is not set. Please set the CUSTOM_ASYNC_TRANSCRIPTION_URL environment variable to the URL of your custom async transcription service."})
 
+        sarvam_settings = value.get("sarvam", {})
```
noah-duncan, line 1176: Hi @RohanDebroy I think this kind of logic should be enforced by Sarvam, not us, so it can be removed. 

If Sarvam's server rejects it, Attendee will show an error message to the user in this screen:

Image

When I tested with these options `{"model":"saarika:v2.5","mode":"translate"}`, Sarvam did not respond with an error.

Part of me also wants to say we should replace those string enums with strings, so that when Sarvam introduces a new model we don't need to update our code. But that is not a big of a deal.

bots/serializers.py
```diff
@@ -1394,6 +1403,10 @@ def validate_transcription_settings(self, value):
         if value.get("deepgram", {}).get("callback") and value.get("deepgram", {}).get("detect_language"):
             raise serializers.ValidationError({"transcription_settings": "Language detection is not supported for streaming transcription. Please pass language='multi' instead of detect_language=true."})
 
+        sarvam_settings = value.get("sarvam", {})
```
noah-duncan, line 1406: Same as above comment, this logic can be removed.

noah-duncan (changes requested):
Thanks @RohanDebroy looks good, just had a small request to remove some logic that I don't think is our responsibility.

Also you'll see that a test is failing because you need to update version.json

RohanDebroy:
Closing in favour of this pr: https://github.com/attendee-labs/attendee/pull/856

bots/serializers.py
RohanDebroy: As discussed, raised a new pr and removed the enum validation altogether here: https://github.com/attendee-labs/attendee/pull/856

PR #848, opened 2026-06-02
perf: batch delete_old_audio_chunks to avoid OOMing

rdoria1:
## Summary
- Rewrite \`delete_old_audio_chunks\` to mirror the batched pattern used by \`delete_old_bot_transcripts\` (#833 + #845).
- Filter recording candidates via \`EXISTS(AudioChunk…)\` so re-runs no-op over already-cleaned recordings instead of re-scanning the whole population.
- Process deletes in fixed-size batches keyed by recording id (default \`--batch-size 100\`), one short transaction per batch — never materialize the full QuerySet in memory.

## Why
The previous implementation looped per recording:
\`\`\`python
old_recordings = Recording.objects.filter(...).distinct()
for recording in old_recordings:
    recording.audio_chunks.all().delete()
\`\`\`

When the candidate set grows to millions of recordings, the QuerySet materialization alone allocates GBs of Python objects. Under memory-constrained cron pods, each scheduled run is OOM-killed before any rows are deleted.

After this change the steady-state daily run only touches the small slice of recordings that crossed the retention boundary since the previous run, and the per-batch memory footprint is constant regardless of population size.

## Behavior
- Daily cron at the configured retention: processes only the new slice; no-ops over already-cleared recordings via the \`EXISTS\` filter.
- Manual backfill at larger \`--days\` values: same batched behavior, just over a larger candidate set; safe to interrupt and resume (idempotent — already-cleared recordings become no-op batches on re-run).
- \`Recording\`, \`Bot\`, \`Utterance\`, etc. remain untouched — same scope as before, only \`AudioChunk\` rows are deleted.

## Test plan
- [ ] Local: run \`--dry-run\` on a fixture with mixed cleared/uncleared recordings; confirm counts match the uncleared subset only.
- [ ] Dev: run for real on dev environment; verify audio chunks for recordings older than retention are 0 afterward.
- [ ] Production cron: monitor first scheduled run after this lands — should now produce a \`LastSuccessfulTime\` on the CronJob, and the dead-tuple counter on \`bots_audiochunk\` should jump from the historical backlog finally getting processed.


noah-duncan (dismissed):
Thanks @rdoria1

rdoria1:
unifying into a single script actually.

noah-duncan (approved):
Thanks @rdoria1

PR #849, opened 2026-06-02
perf: add cleanup_old_data + share bounded cleanup logic across commands

rdoria1:
## Summary

Adds a new \`cleanup_old_data\` management command that handles cleanup for \`Utterance\`, \`AudioChunk\`, and \`BotResourceSnapshot\` from a single entry point, so the deployment side can run **one** CronJob instead of one per table. Adds cleanup for \`bots_botresourcesnapshot\` (currently has no retention and grows unbounded).

The existing per-table commands (\`delete_old_bot_transcripts\`, \`delete_old_audio_chunks\`) are **kept** for backward compatibility — existing CronJobs wired to them continue to work.

To avoid duplicating the deletion logic across three commands, the actual SELECT-and-DELETE loop now lives in a single shared module, \`bots/cleanup.py\`. Each command is a thin wrapper.

### New

- **\`bots/cleanup.py\`** — three pure functions, single source of truth for the bounded deletion logic:
  - \`cleanup_old_utterances(cutoff, batch_size, dry_run)\`
  - \`cleanup_old_audio_chunks(cutoff, batch_size, dry_run)\`
  - \`cleanup_old_bot_resource_snapshots(cutoff, batch_size, dry_run)\`
- **\`bots/management/commands/cleanup_old_data.py\`** — thin dispatcher; parses args, looks up the table → function map, runs each in its own try/except.

### Modified (slimmed to thin wrappers)

- **\`delete_old_bot_transcripts.py\`** — now ~35 lines (was ~80). Args unchanged; body calls \`cleanup_old_utterances\`. **Semantic change**: filters on \`Utterance.created_at\` directly (see *Filtering on the row's own created_at* below).
- **\`delete_old_audio_chunks.py\`** — now ~35 lines (was ~50). Args unchanged; body calls \`cleanup_old_audio_chunks\`. **Semantic change**: filters on \`AudioChunk.created_at\` directly (see below). Default \`--batch-size\` drops from 1000 (post-#848) back to 100 (matching transcripts).

## Filtering on the row's own \`created_at\`

All three functions filter by the target row's **own** \`created_at\`, not by a parent's age (e.g. \`Recording.created_at\` for audio chunks, or the parent bot's ENDED-event timestamp for utterances).

**Why:** scheduled bots can have a parent \`Recording\` created weeks before the bot actually runs and starts producing chunks/utterances. Filtering by the parent's age would over-delete in that case — dropping data that is actually recent.

**Trade-off:** when the cutoff falls mid-recording (or mid-meeting), some of the recording's later chunks (or some of the meeting's later utterances) survive the current run and get cleaned up by the next cron tick. With a reasonable retention buffer this is harmless.

## Boundedness — all three functions

The previous \`delete_old_audio_chunks\` implementation could OOM under memory-constrained pods because the SELECT materialized the full candidate QuerySet into Python memory before the DELETE. \`delete_old_bot_transcripts\` had the same risk in latent form (\`list(BotEvent...)\` + \`recording__bot_id__in=batch\` could DELETE millions of rows per iteration if any bot had many utterances).

The shared pattern, applied uniformly:

1. **Keyset by the target table's \`id\`** (\`id__gt=last_id\` ordered by id, \`LIMIT batch_size\`) — SELECT loads exactly \`batch_size\` integer IDs per iteration regardless of total candidate population.
2. **DELETE by \`id__in=<those ids>\`** — each transaction deletes exactly \`batch_size\` rows, independent of any per-parent fan-out.

Memory footprint and per-transaction DELETE size are now independent of table size for **all three** commands.

## Per-table predicate

Each function isolates its own age filter — they're all now one-line predicates on the row's own column:

| Table | Predicate |
|---|---|
| utterances | \`Utterance.id > last_id AND created_at < cutoff\` |
| audio_chunks | \`AudioChunk.id > last_id AND created_at < cutoff\` |
| snapshots | \`BotResourceSnapshot.id > last_id AND created_at < cutoff\` |

No joins, no subqueries — just a keyset cursor + a column filter, indexed naturally by the PK.

## Args (\`cleanup_old_data\`)

| Flag | Default | Purpose |
|---|---|---|
| \`--days\` | \`30\` | Retention threshold; cron should pass explicitly |
| \`--batch-size\` | \`100\` | Rows deleted per transaction |
| \`--dry-run\` | off | Report counts without deleting |
| \`--tables\` | all | CSV subset of \`utterances,audio_chunks,snapshots\` |

Each block runs inside its own \`try/except\` so a failure in one table doesn't stop the others; the cron retries the failed one next tick.

## Tests

A new \`bots/tests/test_cleanup.py\` covers all three helpers and all three commands. For each helper: old/new mix (only old is deleted), \`dry_run\` reports the count without touching rows, empty table returns 0 cleanly, keyset loops to drain a backlog larger than \`batch_size\`, and re-running on an already-clean table is a no-op. For \`cleanup_old_data\`: \`--tables=utterances\` only touches that table, default runs all three, \`--tables=bogus\` raises \`CommandError\`, \`--dry-run\` doesn't delete, and a failure in one block doesn't stop the others (mocked via \`patch.dict\` on \`TABLE_TO_CLEANUP\`). Plus a smoke test per legacy command confirming the thin wrapper delegates correctly.

## Manual test plan

- [ ] \`python manage.py cleanup_old_data --dry-run\` reports counts for all three tables with no DB writes
- [ ] \`--tables utterances --dry-run\` reports only utterances count
- [ ] \`--tables bogus\` raises a clear \`CommandError\`
- [ ] \`python manage.py delete_old_bot_transcripts --dry-run\` still works (backward compat)
- [ ] \`python manage.py delete_old_audio_chunks --dry-run\` still works (backward compat)
- [ ] Idempotency: run twice in a row, second run finds 0 work to do

## Adding a new table

1. Add a new function to \`bots/cleanup.py\` following the same keyset pattern
2. Add an entry to \`TABLE_TO_CLEANUP\` in \`cleanup_old_data.py\`
3. (Optional) Add a per-table \`delete_old_*\` command as a thin wrapper

That's it — no fan-out plumbing across multiple files.

## Backlog handling

For deployments with a large historical backlog (e.g. data accumulated while a previous cron was failing), the recommended first-time rollout is to suspend the cron, run \`cleanup_old_data --tables audio_chunks --days 90\` (then 60, then 30) manually in detached mode to chew through history in stages, then enable the cron at steady state. Each manual run is bounded the same way the cron is and is safe to interrupt.


pylead (dismissed):

noah-duncan (approved):
Thanks @rdoria1

PR #850, opened 2026-06-03
fix: set Google Meet SSO cookie Secure flag from request scheme

rafaeljusto:
`GoogleMeetSetCookieView` hard-coded `secure=True` on the `google_meet_sign_in_session_id` cookie. In-cluster bot pods reach this view over plain HTTP via `INTERNAL_SITE_DOMAIN`, and browsers silently drop a Secure cookie set over HTTP. The session cookie, therefore, never persisted. `GoogleMeetSignInView` found no cookie on the sign-in hop and the SAML assertion was never built.

This PR gates the Secure flag behind a new `USE_SECURE_COOKIE_FOR_SIGNED_IN_GOOGLE_MEET_BOTS` environment variable (`secure=os.getenv("USE_SECURE_COOKIE_FOR_SIGNED_IN_GOOGLE_MEET_BOTS", "true") == "true"`). 

It defaults to `true`, so the public HTTPS path keeps its Secure cookie with no regression. Deployments that route bots over the internal HTTP path can set it to `false` so the cookie persists and the sign-in flow completes.

Related to:
https://github.com/attendee-labs/attendee/pull/842


noah-duncan (dismissed):
Thanks @rafaeljusto, I'm slightly worried about the `request.is_secure()` check failing because of a configuration where the request is HTTPS, but when it hits the django server, it went through a layer that converted it to HTTP, so it does the wrong thing. So just did the dumb thing and used an env var.

PR #859, opened 2026-06-08
fix: make bot public IP fetch more tolerant and quieter

rdoria1:
## What

The public IP fetch for bot resource snapshots (`get_public_ip` in `bot_resource_snapshot_taker.py`) used a **0.5s timeout** and logged failures at **ERROR** level, producing log lines like:

```
Error getting public IP for bot <id>: <urlopen error timed out>. Continuing...
```

## Why it's harmless

- The fetch runs on a **non-blocking daemon thread**, so it never delays the bot.
- The public IP is **optional, display-only metadata** — it's attached to the first resource snapshot and shown as the "External IP" field on the bot detail page. Nothing branches on it; a miss just omits that one field.

So the 0.5s timeout made failures both common and noisy, despite being non-fatal.

## Change

- Raise the timeout from `0.5s` to `3s` (no blocking cost on the daemon thread, captures the IP far more often).
- Lower the handled-failure log from `error` to `warning`.

No behavior change for bots; reduces log noise and improves IP capture rate.


noah-duncan (approved):
Thanks @rdoria1

PR #861, opened 2026-06-09
fix: make Chrome download visible and fail-fast instead of hanging

rdoria1:
## Problem

The Chrome `.deb` download in the Dockerfile uses `wget -q` with no timeout:

```dockerfile
RUN wget -q https://mirror.cs.uchicago.edu/.../google-chrome-stable_134.0.6998.88-1_amd64.deb
```

With `-q` there is no log output, and with no timeout a stalled connection from a CI runner hangs the build step indefinitely (observed 40+ min vs the usual ~7-8 min) instead of failing or recovering.

## Change

```dockerfile
RUN wget --progress=dot:giga --timeout=30 --tries=3 https://mirror.cs.uchicago.edu/.../google-chrome-stable_134.0.6998.88-1_amd64.deb
```

- **`--progress=dot:giga`** — replaces `-q` so download progress (%, speed) is visible in CI logs. Dot mode is used because the default progress bar needs a TTY and won't render in CI; `giga` scale (~1 MB/dot) keeps the output compact for the ~109 MB file.
- **`--timeout=30 --tries=3`** — a stalled connection now hits an idle timeout, drops, and retries with a fresh connection instead of hanging.

Same URL, same pinned version, same sha256 verification — only the download behavior changes. No functional change to the image.

## Scope / follow-up

This is a **resilience + observability** improvement, not a root-cause fix. The underlying fragility is depending on a single mirror that can be slow/unreachable from CI runners. `--timeout` is an idle timeout, so it catches hard stalls and (via `--tries`) recovers from them, but it will not abort a slow-but-progressing trickle. A fuller fix (a more reliable source, or a `curl` total/min-speed limit) can follow if needed.


noah-duncan (approved):
Thanks @rdoria1

PR #872, opened 2026-06-19
feat: capture pod diagnostics when a bot fails to launch

dihau123:
When a bot pod is created but its container never starts, the bot never heartbeats and the cleanup job eventually marks it FATAL_ERROR / bot_not_launched — with no record of why. The Kubernetes events expire (~1h) before the failure is recorded, leaving the event metadata empty and the failure undiagnosable after the fact.

Capture the pod status and recent Kubernetes events into the bot_not_launched event metadata at cleanup time. The container waiting reason (e.g. ImagePullBackOff, CreateContainerError) persists on the pod object even after events expire, so this still surfaces the root cause. Gathering is best-effort and never raises, so it cannot block the bot from being marked failed.

Add a shared k8s_diagnostics helper and tests for the captured-metadata and pod-already-gone paths.


noah-duncan (dismissed):
Thanks @dihau123

PR #878, opened 2026-06-23
feat: add ECS RunTask bot launch method

jonccrawley:
## Summary
Adds `LAUNCH_BOT_METHOD=ecs`, which launches each bot as a standalone, run-to-completion **ECS task** (one task per bot) — the ECS analogue of the existing Kubernetes `BotPodCreator`. A bot task runs `python manage.py run_bot --botid <id>` and exits when the meeting ends, so there's no service or restart policy to manage. Every deployment value is read from the environment, so the launcher itself is generic.

## What's included
- **`bots/bot_ecs_task_creator/`** — `BotEcsTaskCreator`: `ecs:RunTask` with command + `IS_A_BOT_POD` overrides, FARGATE or capacity-provider, retry on capacity-race / throttling (both the exception and `failures[]` forms), and `find`/`stop` helpers keyed on `startedBy=bot-<id>`.
- **`launch_bot_utils.py`** — dispatch the `ecs` method (same `FATAL_ERROR / BOT_NOT_LAUNCHED` event as the K8s path on failure).
- **`restart_bot_pod_task.py` / `run_correct_failed_bot_launches.py`** — recovery paths made method-aware; **Kubernetes behavior is unchanged**.
- Docs (`docs/environment-variables.md`) + tests (creator incl. retry/error paths, and the ecs dispatch branch).

## Environment contract
`BOT_ECS_CLUSTER`, `BOT_TASK_DEFINITION` (+ `BOT_TASK_DEFINITION_<TYPE>`), `BOT_TASK_SUBNETS`, `BOT_TASK_SECURITY_GROUPS`, `BOT_TASK_LAUNCH_TYPE`/`BOT_TASK_CAPACITY_PROVIDER`, `BOT_TASK_ASSIGN_PUBLIC_IP`, `BOT_TASK_CONTAINER_NAME` (default `bot-proc`). The bot task definition carries the full app env/secrets, exactly like the K8s configmap+secret.

## Notes
- The voice-agent **webpage-streamer** sidecar isn't supported in `ecs` mode yet (logs a warning and launches in recording/transcription mode).
- Unlike K8s (deterministic, reused pod names), ECS task IDs are never reused, so the stop-and-wait in `restart_bot_pod` is a clean-restart safety measure, not a correctness requirement.
- Validated end-to-end on a self-hosted ECS/Fargate deployment (Google Meet bot joined, recorded, transcribed).


noah-duncan:
Thanks @jonccrawley  this is cool stuff. People who don't want to deal with k8s will appreciate it. If you look at this [part of the code](https://github.com/attendee-labs/attendee/blob/2476308b1cd7903644ad683f756d9a461ade4bea/bots/management/commands/clean_up_bots_with_heartbeat_timeout_or_that_never_launched.py#L64) you'll see we have a section where we terminate bot infra that is running, but got into a messed up state where it isn't emitting a heartbeat. You probably want to add something for terminating the ECS container.

And yea, at some point we'll refactor the code to get rid of the ugly conditional logic like this
```
if os.getenv("LAUNCH_BOT_METHOD") == x
elsif os.getenv("LAUNCH_BOT_METHOD") == y
```

PR #880, opened 2026-06-24
fix: correct virtual display size to 1920x1080 to match mocap sequences

marvellousz:
## Problem
When self-hosting Attendee, the Google Meet bot consistently fails with:

    "Could not find mocap sequence landing on target element after 10 attempts"

The bot reaches the pre-join screen successfully (name input, mic, camera) but cannot click the "Join now" button.

## Root cause
`web_bot_adapter.py` line 688 starts the virtual Xvfb display at `(1930, 1090)`, but the mocap sequences were recorded at 1920×1080. The 10px mismatch shifts UI element positions just enough that every simulated click lands outside the button boundary after 10 retries.

## Fix
Change `Display(visible=0, size=(1930, 1090))` → `Display(visible=0, size=(1920, 1080))`

## Verified
After this fix, the bot successfully clicks "Join now", joins the meeting,
and completes recording on a self-hosted deployment.


noah-duncan:
Hi @marvellousz thanks for the PR. Unfortunately, we need the dimensions to be (1930, 1090) for the screen recording to produce (1920, 1080). 

Are you seeing the bot actually fail to join the meeting or is it just taking longer than it needs to because of retries?

I have an experimental branch [here](https://github.com/attendee-labs/attendee/tree/synthetically-generated-mocap-experiment) that uses generated mocap. Can you see if this eliminates the problem for you?

marvellousz:
Thanks for the context! Makes sense. I will test the synthetically-generated-mocap-experiment branch and report whether it resolves the join failure on self hosted deployments.

marvellousz:
Hey @noah-duncan, Tested the synthetic mocap approach locally, bot joins successfully at the original 1930×1090 display size without any resolution fix needed. The minimum jerk motor model works correctly. Happy to close this PR in favor of the experimental branch landing.

PR #882, opened 2026-06-25
Fix Zoom Web mixed audio websocket forwarding

### LINKED ISSUE #881: Zoom Web mixed audio websocket path never starts
## Problem

Zoom Web bots can detect meeting audio tracks, but the mixed-audio websocket path doesn't start reliably.

In practice this can look like:
- captions still work
- per-participant audio track detection still works
- but `realtime_audio.mixed` never starts for Zoom Web

## Root cause

The Zoom Web payload had mixed-audio frame forwarding logic available, but it did not bootstrap the mixed-audio stream when the first meeting audio track arrived.

That meant the runtime could define the forwarding path without ever initializing it at the right point in the meeting lifecycle.

## Reproduction

1. Launch a Zoom Web bot with `zoom_settings.sdk=web`
2. Configure `websocket_settings.audio`
3. Join a real Zoom meeting with active audio
4. Observe that Zoom track events appear, but external mixed-audio websocket output may never begin

## Expected behavior

When the first meeting audio track arrives, the Zoom Web adapter should initialize the mixed-audio stream and start forwarding `realtime_audio.mixed` chunks.

## Evidence

I verified this against a real Zoom meeting locally and narrowed it to the mixed-audio stream bootstrap point in `bots/zoom_web_bot_adapter/zoom_web_chromedriver_payload.js`.

I also prepared a PR with a minimal fix and regression coverage.

booynal:
## Summary

Fix the Zoom Web mixed-audio websocket path so it starts when the first meeting audio track arrives.

## Problem

Zoom Web bots could detect meeting audio tracks, but mixed audio was not reliably forwarded over the websocket path.

Observed behavior:
- captions could still work
- per-participant audio track detection could still work
- but `realtime_audio.mixed` would never start for Zoom Web

Fixes #881.

## Root cause

The Zoom Web payload defined the mixed-audio forwarding path, but it did not bootstrap the mixed-audio stream when the first meeting audio track arrived.

That meant the runtime could have frame-forwarding code present without ever initializing the mixed stream at the correct lifecycle point.

## Changes

- store the mixed audio track from the mixed destination stream
- start `processMixedAudioTrack()` when mixed audio is enabled and a mixed track exists
- bootstrap the mixed-audio stream from `addAudioTrack()` when the first meeting audio track arrives
- guard bootstrap so it only happens before the mixed stream has been initialized
- add regression coverage for:
  - payload mixed-audio websocket forwarding
  - mixed-audio stream initialization on first track arrival
  - Python-side mixed-audio frame decoding to callback
  - controller websocket payload emission for `realtime_audio.mixed`

## Test Plan

- [x] `python manage.py test bots.tests.test_zoom_web_bot.TestZoomWebBot.test_zoom_web_payload_routes_mixed_audio_frames_to_websocket bots.tests.test_zoom_web_bot.TestZoomWebBot.test_zoom_web_payload_initializes_mixed_audio_stream_when_track_arrives bots.tests.test_zoom_web_bot.TestZoomWebBot.test_zoom_web_process_mixed_audio_frame_forwards_pcm_to_callback bots.tests.test_zoom_web_bot.TestZoomWebBot.test_zoom_web_mixed_audio_callback_sends_realtime_audio_payload`
- [x] Manual validation against a real Zoom meeting with `zoom_settings.sdk=web`
- [x] Verified external websocket client received continuous `realtime_audio.mixed` messages

### DIFF
--- a/bots/tests/test_zoom_web_bot.py
+++ b/bots/tests/test_zoom_web_bot.py
@@ -3,8 +3,11 @@ import os
 import struct
 import threading
 import time
+from base64 import b64decode
+from pathlib import Path
 from unittest.mock import MagicMock, patch
 
+import numpy as np
 import requests
 from django.db import connection
 from django.test import TransactionTestCase, tag
@@ -13,6 +16,7 @@ from selenium.common.exceptions import NoSuchElementException
 from bots.bot_adapter import BotAdapter
 from bots.bot_controller.bot_controller import BotController
 from bots.models import Bot, BotEventManager, BotEventSubTypes, BotEventTypes, BotStates, Credentials, Organization, Project, Recording, RecordingTypes, TranscriptionProviders, TranscriptionTypes, WebhookDeliveryAttempt, WebhookSubscription, WebhookTriggerTypes, ZoomMeetingToZoomOAuthConnectionMapping, ZoomOAuthApp, ZoomOAuthConnection, ZoomOAuthConnectionStates
+from bots.zoom_web_bot_adapter.zoom_web_bot_adapter import ZoomWebBotAdapter
 
 
 # Helper functions for creating mocks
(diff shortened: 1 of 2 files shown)


booynal:
Closing per request.

PR #905, opened 2026-07-13
Add scheduler heartbeat file for liveness probing

rdoria1:
## What

Have `run_scheduler` write a heartbeat file at the end of every polling cycle (and once at startup), so a Kubernetes liveness probe can detect when the loop has stalled.

## Why

`run_scheduler` runs as a single long-lived process. Its loop makes Redis and DB calls that can block indefinitely (no socket/statement timeouts), so if one of those calls hangs, the process stays **alive but frozen** — it stops launching scheduled bots without crashing or raising. A plain process-liveness check can't tell this apart from a healthy process, so nothing restarts it and scheduled bots silently stop launching until someone redeploys.

Writing a heartbeat file each cycle turns this into something an exec liveness probe can catch: if the file goes stale, the loop is stuck and the pod can be restarted automatically.

## Changes

- `SCHEDULER_HEARTBEAT_FILE` env var (default `/tmp/scheduler_heartbeat`).
- `_write_heartbeat()` rewrites the file with the current timestamp; failures are logged and swallowed so heartbeat I/O can never take the scheduler down.
- Called once at startup and at the end of each cycle (after the `finally` cleanup), so a mid-cycle hang leaves the file stale rather than refreshing it.
- Two unit tests covering the write and the error-swallowing path.

This change is inert on its own — it only writes a file. Wiring up the liveness probe that reads it is done in deployment config.

### DIFF
--- a/bots/management/commands/run_scheduler.py
+++ b/bots/management/commands/run_scheduler.py
@@ -25,6 +25,10 @@ log = logging.getLogger(__name__)
 
 CALENDAR_SYNC_THRESHOLD_HOURS = 24  # The longest a calendar can go without having been synced
 
+# Heartbeat file the scheduler rewrites each cycle so a liveness probe can
+# restart the pod if the loop stalls (a process-liveness check wouldn't catch it).
+SCHEDULER_HEARTBEAT_FILE = os.getenv("SCHEDULER_HEARTBEAT_FILE", "/tmp/scheduler_heartbeat")
+
 
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


noah-duncan (changes requested):
Thanks @rdoria1 this looks reasonable. But I don't think this behavior should immediately start happening for everyone by default. Could you make the SCHEDULER_HEARTBEAT_FILE env var default to None and add a line at the start of _write_heartbeat that aborts if it is undefined?

PR #908, opened 2026-07-14
build: drop kernel headers and apply security updates in runtime image

GeorgelPreput:
## Summary

The image ships build-only packages in the final runtime stage. The largest offender is **`linux-libc-dev`** (Linux kernel *headers*), which is installed twice (once in the main dependency block, again in a dedicated `# Install Linux Kernel Dev` step) and is only needed to compile native Python extensions (`av`, opencv) at build time. At runtime the container uses the **host kernel**, so these headers are dead weight.

A Trivy scan of the `ubuntu:22.04`-based image attributes **~2,500 of ~4,000** OS-package CVEs to `linux-libc-dev` alone — including **every CRITICAL** and the **vast majority of HIGH** findings. None are reachable at runtime: already-compiled `.so` files link the runtime `libc` (`libc6`), not the headers.

Separately, the build never runs `apt-get upgrade`, so a number of CVEs that Ubuntu has **already patched** (openssh-client, libsoup2.4, gzip, ncurses, libexif, libidn12, mesa) linger at outdated versions.

## Change

- Remove the redundant second `linux-libc-dev` install (it stays in the main block, so it's still available while native extensions compile).
- In a final step in the `build` stage — after all compilation is done — purge `linux-libc-dev` (and its build-only dependents) and apply pending Ubuntu security updates.

This takes the image from **177 → 2** HIGH/CRITICAL Trivy findings and clears all CRITICALs, with no change to runtime behaviour.

## Safety

Verified locally that `apt-get purge -y linux-libc-dev` on `ubuntu:22.04` removes only build-only packages — `build-essential`, `g++`, `g++-11`, `libc6-dev`, `libstdc++-11-dev`, `linux-libc-dev` — and **retains the runtime `libc6`**. Chrome is installed from a pinned local `.deb` with no apt source, so `apt-get upgrade` does not touch it.

## Note

A longer-term improvement would be a build/runtime multi-stage split to also drop `build-essential`, `cmake`, `gfortran`, `gdb`, `git`, `vim` and the `*-dev` headers from the runtime image. This PR is the minimal, low-risk change that keeps the existing single-stage layout.


noah-duncan (changes requested):
Hi @GeorgelPreput appreciate the changes, this is a good idea. However it is breaking all of the ci tests. I think you removed something that was actually needed.

GeorgelPreput:
Yikes! Sorry, I'll have a look at the CI results tonight and try to fix. Let me get back to you on this.

GeorgelPreput:
@noah-duncan thanks for re-running the tests, I think I was too aggressive at first with dropping the dependencies. Some things that are needed at runtime were actually second-hand dependencies from the dev packages.

I do notice however there's a requested change that `--platform=linux/amd64` not be hardcoded in the Dockerfile. That's pre-existing before my changes, appears on `main` as well. Shall I remove that, or can it stay? (as a btw, some of the packages you're using are genuinely AMD64-only, so this Docker cannot be built on ARM in its current state anyway)

PR #922, opened 2026-07-23
fix: support trusted CSRF origins in production

booynal:
## Summary
- add a comma-separated `CSRF_TRUSTED_ORIGINS` setting to the standard production configuration
- trim entries and ignore blank values
- bump the patch version to 1.58.6

## Why
Self-hosted deployments using `attendee.settings.production` behind an HTTPS reverse proxy can otherwise reject cookie-authenticated unsafe requests from configured application origins with Django CSRF 403 errors.

## Testing
- `python -m py_compile attendee/settings/production.py`
- `python -m json.tool version.json`
- `git diff --check github/main`

Full Django tests were not run locally because this host does not have the project dependencies installed.


noah-duncan:
Fixed here: https://github.com/attendee-labs/attendee/pull/952

PR #925, opened 2026-07-27
fix: Connect late-arriving Google Meet audio tracks to the mixed-audio graph

GriffinCanCode:
## Problem

`StyleManager` builds the Google Meet meeting-audio mix exactly once, in `startSilenceDetection()` — it snapshots `this.audioTracks` into source nodes and connects those to a `MediaStreamDestination`:

```js
this.audioSources = this.audioTracks.map(track => {
    const mediaStream = new MediaStream([track]);
    return this.audioContext.createMediaStreamSource(mediaStream);
});
const destination = this.audioContext.createMediaStreamDestination();
this.audioSources.forEach(source => source.connect(destination));
```

`addAudioTrack()` keeps collecting every audio track the peer connection negotiates afterwards, but only pushes to the array:

```js
addAudioTrack(audioTrack) {
    this.audioTracks.push(audioTrack);
}
```

Nothing ever connects those later tracks to `destination`. The mix carries only the tracks that happened to exist at graph-build time, and is entirely silent when that set was empty.

The ordering makes this the common case rather than an edge case. `addAudioTrack()` is called from the `track` event on the meeting peer connection, whenever Meet negotiates one. `startSilenceDetection()` runs from `start()`, which runs from `enableMediaSending()`. So any participant whose audio track is negotiated after the bot starts sending media is inaudible in the mix — as is anything Meet re-negotiates while the bot is in the call.

## Impact

Three consumers read that one graph, so they all observe the same silence:

- `getMeetingAudioStream()` — the stream `shared_chromedriver_payload.js` adds to the webpage-streamer peer connection, which becomes the voice agent page's `getUserMedia` microphone. A voice agent in Google Meet hears nothing and never responds.
- `processMixedAudioTrack()` — the `sendMixedAudio` websocket output, taken from `destination.stream.getAudioTracks()[0]`.
- `checkAudioActivity()` — the silence detector, whose analyser is fed by `createMediaStreamSource(destination.stream)`. A meeting with real audio can read as silent.

## Fix

`MixedAudioStreamManager` in the Teams adapter already solves this — it patches a late track straight into the live destination and only queues one when the graph doesn't exist yet:

```js
// If start() already ran, patch the new track into the existing mix.
if (this.audioContext && this.destination) {
    const source = this.audioContext.createMediaStreamSource(mediaStream);
    source.connect(this.destination);
    this.sourceNodes.push(source);
}
else {
    this.audioTracksToBeAdded.push(track);
}
```

This applies the same idiom to Google Meet: hold the destination on `this.mixDestination`, and have `addAudioTrack()` connect into it when the graph is already live. Tracks that arrive before it exists are unaffected — they are still picked up out of `this.audioTracks` when the mix is built. Adding a source to a `MediaStreamDestination` after the fact is reflected in the already-handed-out `destination.stream`, so no consumer needs to re-fetch anything.

+11 lines, no behavior change on the pre-existing path.

## Testing

There is no JS harness for the injected payloads, so this was verified by running it: with the patch, a Google Meet voice agent that previously received a flat near-silent signal receives the participants' audio and responds normally. Before the patch, the mixed stream measured at effectively the noise floor for the whole call regardless of who was speaking.

`node --check` passes on the modified payload.

### DIFF
--- a/bots/google_meet_bot_adapter/google_meet_chromedriver_payload.js
+++ b/bots/google_meet_bot_adapter/google_meet_chromedriver_payload.js
@@ -232,6 +232,7 @@ class StyleManager {
 
         this.audioContext = null;
         this.audioTracks = [];
+        this.mixDestination = null;
         this.silenceThreshold = 0.5;
         this.silenceCheckInterval = null;
         this.memoryUsageCheckInterval = null;
@@ -243,6 +244,15 @@ class StyleManager {
 
     addAudioTrack(audioTrack) {
         this.audioTracks.push(audioTrack);
+        // If startSilenceDetection() already ran, patch the new track into the existing
+        // mix. Without this, every track Meet negotiates after the mix is built is
+        // collected here and never heard. Tracks added before it exists are picked up
+        // out of this.audioTracks when the mix is built.
+        if (this.audioContext && this.mixDestination) {
+            const source = this.audioContext.createMediaStreamSource(new MediaStream([audioTrack]));
+            source.connect(this.mixDestination);
+            this.audioSources.push(source);
+        }
     }
 
     checkAudioActivity() {
@@ -330,6 +340,7 @@ class StyleManager {
  
          // Create a destination node
          const destination = this.audioContext.createMediaStreamDestination();
+         this.mixDestination = destination;
  
          // Connect all sources to the destination
          this.audioSources.forEach(source => {


GriffinCanCode:
Our in-house agent discovered this bug and had to patch it ourselves. Hopefully this is useful for you guys. Apologies if it's not up to your standards for submitted PRs. We did this quickly.

noah-duncan:
@GriffinCanCode Thanks for the PR. Just want to understand the severity of the issue. Have you ever observed late-arriving audio tracks? 

AFAIK Google Meet creates 3 audio tracks at the beginning and these audio tracks are never replaced or updated.

GriffinCanCode:
Yes @noah-duncan we hit it on every Google Meet join, and it's not intermittent.

I think the "3 tracks created at the beginning, never replaced" model is right; the
bug doesn't depend on tracks being replaced or updated, only on when they're
negotiated relative to when the mix graph is built.

addAudioTrack() is driven by the `track` event on the meeting peer connection
(payload L2116). startSilenceDetection() runs from StyleManager.start() <-
enableMediaSending() <- after_bot_can_record_meeting(), i.e. on
`RecordingPermissionChange: granted` (web_bot_adapter.py L421-423, L837-844). 

Basically the graph is frozen at the instant the bot is admitted, which is the earliest moment it's
in the call, its not after the room has settled. Any track Meet negotiates to that PC
after admission is collected into this.audioTracks and never connected.

The case that bit us is a voice agent that joins ahead of the humans (or is admitted
out of the lobby): at snapshot time there is literally nobody to have a track, so the
destination gets zero sources and stays silent for the whole call.

Measured, on an unpatched build: a 10-minute Meet call where the mixed stream sat at
the noise floor (RMS ~4e-4) the entire time regardless of who was speaking, yielding
7 characters of transcript. Same setup with this patch: the agent receives the
participants' audio and responds normally. Attendee's own silence detector reads the
same graph, so it agreed the meeting was silent.

Worth noting the patch is a no-op under your model: if all the tracks really do exist
before the graph is built, the `if (this.audioContext && this.mixDestination)` branch
never fires on a fresh join and nothing changes. It only adds a path for tracks that
arrive after — which is exactly what MixedAudioStreamManager already does on the
Teams side.

Happy to add a temporary ws log of this.audioTracks.length at snapshot time and post
a capture from a live join if that would settle it faster.

noah-duncan:
Hey @GriffinCanCode thanks for the details. I'm not able to reproduce your issue even when the bot joins an open Google Meet before any of the humans join. Like you said there's nothing wrong with your PR, so we can merge it, but it would be good for me to be able to reproduce this to fully understand the root cause.

If you wouldn't mind could you DM me a meeting link that has this issue on our [community slack](https://join.slack.com/t/attendee-community/shared_invite/zt-3l43ns8cl-G8YnMccWVTugMlloUtSf9g)?

GriffinCanCode:
@noah-duncan that makes sense, and I am pretty sure its because the recording path won't show this.

screen_and_audio_recorder.py captures with `ffmpeg -f x11grab -i $DISPLAY -f alsa -i
default`, i.e. the X display plus whatever Chrome plays out its speakers. That's
completely independent of the WebAudio graph startSilenceDetection() builds, so a
recording sounds perfect even when the mix is dead. 

Only three consumers read that graph:

  - getMeetingAudioStream() -> the webpage-streamer PC -> the voice agent page's
    getUserMedia mic  (this is the one that bit us)
  - processMixedAudioTrack() -> the sendMixedAudio websocket output
  - checkAudioActivity() -> the silence detector

So the meeting itself isn't the variable and a link won't add much — any Meet does it.
The variable is the bot config. To reproduce, create a bot with voice_agent_settings.url
set (or websocket audio with sendMixedAudio), have it join before any human, then have
someone join and talk. Unpatched, the mixed stream stays at the noise floor for the rest
of the call.

Cheapest hard proof, no voice agent needed - two log lines in the payload:

  addAudioTrack(audioTrack) {
      window.ws?.sendJson({type: 'Dbg', at: 'addAudioTrack',
                           t: performance.now(), live: !!this.mixDestination});
      ...
  }

  startSilenceDetection() {
      window.ws?.sendJson({type: 'Dbg', at: 'snapshot',
                           t: performance.now(), n: this.audioTracks.length});
      ...
  }

Any addAudioTrack line printing after the snapshot line is a track that is inaudible in
the mix by construction. If in your environment n is already 3 at snapshot and nothing
arrives after, that's a real difference between our setups worth understanding — we're
self-hosted, and the ordering here is a race between Meet negotiating the downstream
audio transceivers and `RecordingPermissionChange: granted` firing, so RTT and lobby vs
open-join could plausibly flip it.

Happy to DM on Slack either way - I can grab a capture off a live join with those two
lines in if that's the fastest path for you.

noah-duncan:
Hi @GriffinCanCode I understand what you're saying about the screen recorder being separate from the web audio graph. I made sure to test the web audio code. I tested with a voice agent in the attempt to reproduce mentioned above. The voice agent worked as expected and it joined before any humans. Would appreciate the DM of your meeting so I can reproduce this.

noah-duncan:
This is what the bot logs when joining an open google meet before anyone else joins.
```
Received JSON message: {'type': 'Dbg', 'at': 'addAudioTrack', 't': 5354.800000000745}
Received JSON message: {'type': 'Dbg', 'at': 'addAudioTrack', 't': 5355.700000001118}
Received JSON message: {'type': 'Dbg', 'at': 'addAudioTrack', 't': 5356.10000000149}
Received JSON message: {'type': 'Dbg', 'at': 'snapshot', 't': 19551, 'n': 3}
```

GriffinCanCode:
Hey @noah-duncan that's a clean result and it changes my picture, thanks for actually running it. A
14-second margin between the tracks and the snapshot is not a race, so for a bot walking
into an open meeting your model is right and there's nothing to reproduce there.

I think the case that bites us is later than join. RTCInterceptor attaches the `track`
listener per peer connection, so if Meet ever recreates the PC mid-call — network blip,
ICE restart, SFU rebalance — a fresh batch of addAudioTrack fires long after the
snapshot, the previous tracks end, and the mix goes from working to dead for the rest
of the meeting. That fits what we see better than the join ordering does: not a bot that
starts deaf, but one that goes deaf and stays deaf, with the silence detector agreeing.

Two things that would test it faster than a link:

  - Leave your instrumented bot in a longer call and drop its container's network for
    ~5s mid-meeting. If a second batch of addAudioTrack lines appears with live=false,
    that's the failure without needing our environment at all.
  - Same run, add a screen share and have a 4th and 5th participant talk — anything that
    makes Meet negotiate beyond the initial 3.

I sadly am not in a place I can join a meet right now, I guess I can send you a meet link for later not sure how much help that would be realistically. But I'm also going to run the same two log lines on
our side and post the timeline here — if our snapshot shows n=3 with nothing arriving
late, then our silence had a different root cause and this PR is a correctness
hardening rather than the fix for our bug, and I'd rather say so than have you chase it.

noah-duncan:
Thanks would appreciate a willingness to spend more time digging into this. I don't have anything against LLMs but the analysis here seems like it was heavily delegated to an LLM and it's possible the LLM is operating under some misconceptions.

You likely did find a genuine issue but want to be sure that the fix actually fixes the issue.

PR #929, opened 2026-07-29
fix: handle unavailable Zoom participants during reconnect

solaoi:
## Summary

- Handle temporarily unavailable Zoom participant details without crashing the bot.
- Clear the only-participant timer and defer evaluation until a later participant update.
- Add regression coverage for the missing-participant case.
- Bump `version.json` from `1.58.9` to `1.58.10` for the patch release.

## Root cause

During Zoom native reconnect or waiting-room transitions, `GetParticipantsList` can expose a participant ID before `GetUserByUserID` can resolve it. For an uncached participant, `get_participant()` then returns `None`. `update_only_one_participant_in_meeting_at()` dereferenced that value, and the resulting Python exception escaped through the Zoom SDK nanobind callback, terminating the process with `SIGABRT`. The bot was later reported as a heartbeat timeout.

## Impact

Zoom native bots no longer terminate when the SDK participant list is transiently inconsistent. Resetting the timer also avoids making an automatic-leave decision from incomplete participant data.

## Validation

- `python manage.py test bots.tests.test_zoom_bot.TestZoomParticipantTracking --verbosity 2`
- `ruff check --no-cache bots/zoom_bot_adapter/zoom_bot_adapter.py bots/tests/test_zoom_bot.py`
- `ruff format --no-cache --check --diff bots/zoom_bot_adapter/zoom_bot_adapter.py bots/tests/test_zoom_bot.py`


bots/zoom_bot_adapter/zoom_bot_adapter.py
```diff
@@ -271,6 +271,10 @@ def update_only_one_participant_in_meeting_at(self):
         other_bots_in_meeting_names = []
         for participant_id in all_participant_ids:
             participant = self.get_participant(participant_id)
+            if participant is None:
+                logger.warning(f"Unable to update only-participant timer because participant {participant_id} is not available")
+                self.only_one_participant_in_meeting_at = None
```
noah-duncan, line 275: Thanks this is a good approach. I was thinking about suggesting you sleep for a really brief period and retry `self.get_participant(participant_id)`, but this is simpler.

In the worst case we will fail to register that the new participant that joined was a bot and incorrectly reset `only_one_participant_in_meeting_at`, but this happens quite rarely and the silence timeout would catch it.

bots/zoom_bot_adapter/zoom_bot_adapter.py
```diff
@@ -271,6 +271,10 @@ def update_only_one_participant_in_meeting_at(self):
         other_bots_in_meeting_names = []
         for participant_id in all_participant_ids:
             participant = self.get_participant(participant_id)
+            if participant is None:
+                logger.warning(f"Unable to update only-participant timer because participant {participant_id} is not available")
```
noah-duncan: Could you change this to mention that we are setting `only_one_participant_in_meeting_at` to None in order to be consistent with the log statement on line 286?

noah-duncan (changes requested):
Hi @solaoi thanks for the well-written PR. I just had a small request to change the logging statement. After that will be happy to merge.

bots/zoom_bot_adapter/zoom_bot_adapter.py
noah-duncan: BTW, made you a collaborator of the repo, so you no longer need to fork it. Once you open PRs without forking, our CI suite will run on your PR automatically.

solaoi: Updated in 05da658d: the warning now explicitly says that only_one_participant_in_meeting_at was set to None, and the regression assertion checks the new message. Thanks for adding me as a collaborator as well; I will use direct branches for future PRs.

bots/zoom_bot_adapter/zoom_bot_adapter.py
solaoi: Thanks for confirming. I kept the simpler early-return behavior without adding a retry.

noah-duncan (approved):
Thanks @solaoi !

PR #930, opened 2026-07-29
feat: allow disabling bot transcription

solaoi:
## Problem

Zoom bots using the native SDK implicitly select Deepgram when `transcription_settings` is omitted. This is useful for existing users who rely on the default, but consumers that only need recording or realtime audio streaming currently have no explicit way to disable Attendee transcription. Without Deepgram credentials, those bots produce `credentials_not_found`; with credentials, audio can be sent to a third-party transcription provider even when no Attendee transcript is needed.

Changing the existing default would break users who intentionally rely on the Deepgram fallback. The opt-out is therefore explicit and lives in `transcription_settings` rather than in Zoom-specific settings.

## Changes

- add the generic boolean option `transcription_settings.enabled`
- set the recording provider to `NO_TRANSCRIPTION` when `enabled` is `false`, while leaving the existing transcription type unchanged
- preserve existing platform defaults when `transcription_settings` is omitted
- require an explicit transcription provider when `enabled` is `true`
- reject `enabled: false` when a transcription provider is also configured
- stop forwarding per-participant audio chunks to transcription when transcription is disabled
- keep Zoom native one-way audio reception locally for silence detection, without reading or forwarding its buffer when no callback is configured
- avoid unnecessary Zoom RTMS audio and transcript delivery when their callbacks are absent
- keep platform caption UI behavior unchanged; disabled captions are not collected as Attendee transcripts
- keep recording, mixed-audio capture, and realtime audio WebSocket streaming unchanged
- reject disabling transcription for async transcription requests, where a provider is required
- document the option in the transcription guide and OpenAPI schema

## Example

```json
{
  "zoom_settings": {
    "sdk": "native"
  },
  "transcription_settings": {
    "enabled": false
  }
}
```

## Validation

- `TestCreateBot`: 38 tests passed
- Ruff lint and format checks for changed Python files
- `git diff --check`
- GitHub Actions CI


noah-duncan:
Hi @solaoi , I understand the motivation behind this PR. It is confusing how the Zoom Native bot defaults to deepgram and will generate errors if you don't have deepgram credentials.

However, the Zoom-specific `use_default_transcription` parameter is rather confusing. I think a cleaner solution to this is to allow the user to specify that they don't want the bot to use transcription at all within the `transcription_settings` object.

An alternative solution is using the Zoom Web SDK bot. That will default to using closed captions since we can obtain these in the Web SDK, so you won't get the `credentials_not_found` error messages. However, you may have a reason to prefer the native adapter such as less resource consumption. Pass `"zoom_settings":{"sdk":"web"}` to activate the web bot adapter.

bots/bots_api_utils.py
```diff
@@ -287,8 +289,8 @@ def create_bot(data: dict, source: BotCreationSource, project: Project) -> tuple
             Recording.objects.create(
                 bot=bot,
                 recording_type=bot.recording_type(),
-                transcription_type=TranscriptionTypes.NON_REALTIME,
```
noah-duncan, line 290: The `transcription_type` column is not used anywhere within the business logic. It should probably be removed from the codebase entirely. So I would leave it as is, even though it looks weird to set it to NON_REALTIME.

bots/bot_controller/bot_controller.py
```diff
@@ -109,7 +109,7 @@ def per_participant_audio_input_manager(self):
             return self.per_participant_non_streaming_audio_input_manager
 
     def save_utterances_for_individual_audio_chunks(self):
-        return self.get_recording_transcription_provider() != TranscriptionProviders.CLOSED_CAPTION_FROM_PLATFORM
+        return self.get_recording_transcription_provider() not in (None, TranscriptionProviders.CLOSED_CAPTION_FROM_PLATFORM)
```
noah-duncan: Instead of having `None`, can we have `TranscriptionProviders.NO_TRANSCRIPTION`? That makes it more clear IMO

noah-duncan (changes requested):

solaoi:
Thanks for the feedback — that makes sense.

I plan to replace the Zoom-specific `use_default_transcription` setting with an explicit option inside `transcription_settings`:

```json
{
  "transcription_settings": {
    "no_transcription": {}
  }
}
```

The intended behavior is:

- Omitting `transcription_settings` preserves the existing defaults, including Deepgram for the Zoom Native SDK.
- Explicit providers continue to work as before.
- `no_transcription` skips both external transcription providers and platform captions.
- Combining `no_transcription` with another provider will be rejected as invalid.
- Mixed-audio WebSocket output remains available.

I’ll also address the inline comments by adding `TranscriptionProviders.NO_TRANSCRIPTION` instead of using `None`, while leaving `transcription_type` as `NON_REALTIME`.

Does the `{"no_transcription": {}}` API shape look good to you? If so, I’ll update this PR accordingly.

noah-duncan:
> Does the `{"no_transcription": {}}` API shape look good to you? If so, I’ll update this PR accordingly.

Can you have the attribute be called `enabled` and have it be boolean-valued instead of object-valued? So to disable transcription you would pass `"transcription_settings":{"enabled":false}`

solaoi:
Thanks for the guidance. I’ve updated the PR to use `transcription_settings.enabled` as requested and addressed the inline feedback.

The final behavior is:

- omitting `transcription_settings` preserves the existing platform defaults;
- `{"enabled": false}` disables both third-party transcription and Attendee ingestion of platform captions;
- `enabled: false` with a provider is rejected;
- explicit `enabled: true` requires an explicit provider;
- recording, mixed-audio WebSocket output, and Zoom native silence detection remain unchanged.

I also added `TranscriptionProviders.NO_TRANSCRIPTION`, kept `transcription_type` as `NON_REALTIME`, and added regression coverage. All CI checks are passing.

This should now be ready for re-review. Thanks!

solaoi:
I am closing this because it will be addressed in the following PR.
https://github.com/attendee-labs/attendee/pull/932

PR #931, opened 2026-07-30
Add MinIO service for running the full stack locally

mateomurphy:
Wires an S3-compatible MinIO container into dev.docker-compose.yaml with automatic bucket creation, and fixes path-style addressing (required by MinIO/LocalStack) in both the django-storages config and the raw boto3 upload clients used for recordings.

### DIFF
--- a/attendee/settings/base.py
+++ b/attendee/settings/base.py
@@ -283,6 +283,10 @@ else:
             "secret_key": os.getenv("AWS_SECRET_ACCESS_KEY"),
         },
     }
+    # S3-compatible endpoints like MinIO/LocalStack don't support virtual-hosted-style
+    # requests (bucket-as-subdomain), so fall back to path-style addressing for them.
+    if os.getenv("AWS_ENDPOINT_URL"):
+        DEFAULT_STORAGE_BACKEND["OPTIONS"]["addressing_style"] = "path"
     # Deep copy the DEFAULT_STORAGE_BACKEND
     RECORDING_STORAGE_BACKEND = copy.deepcopy(DEFAULT_STORAGE_BACKEND)
     RECORDING_STORAGE_BACKEND["OPTIONS"]["bucket_name"] = AWS_RECORDING_STORAGE_BUCKET_NAME
--- a/attendee/settings/base.py
+++ b/attendee/settings/base.py
@@ -283,9 +283,11 @@ else:
             "secret_key": os.getenv("AWS_SECRET_ACCESS_KEY"),
         },
     }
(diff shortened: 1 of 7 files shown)


attendee/settings/base.py
```diff
@@ -283,6 +283,10 @@
             "secret_key": os.getenv("AWS_SECRET_ACCESS_KEY"),
         },
     }
+    # S3-compatible endpoints like MinIO/LocalStack don't support virtual-hosted-style
+    # requests (bucket-as-subdomain), so fall back to path-style addressing for them.
+    if os.getenv("AWS_ENDPOINT_URL"):
```
noah-duncan: To avoid altering the behavior for people who have specified `AWS_ENDPOINT_URL` but not need path-style addressing (IE Google Cloud Storage users), I think you should have an env var specifically to turn on path style addressing instead of assuming that if AWS_ENDPOINT_URL is specified, than we want to turn it on.

This also is going to break the experience for people who were running locally but not on MinIO. You need to revise this PR to avoid that.

noah-duncan (changes requested):
Hi @mateomurphy this looks good, but I'd rather not break the experience for everyone else who was running locally, but not using MinIO. Please make the PR backwards-compatible.

init_env.py
```diff
@@ -16,10 +16,11 @@ def main():
 
     print(f"CREDENTIALS_ENCRYPTION_KEY={credentials_key}")
     print(f"DJANGO_SECRET_KEY={django_key}")
-    print("AWS_RECORDING_STORAGE_BUCKET_NAME=")
-    print("AWS_ACCESS_KEY_ID=")
-    print("AWS_SECRET_ACCESS_KEY=")
+    print("AWS_RECORDING_STORAGE_BUCKET_NAME=attendee-recordings")
```
noah-duncan: Please remove these.

dev.docker-compose.yaml
```diff
@@ -11,6 +11,9 @@ services:
       - DJANGO_SETTINGS_MODULE=attendee.settings.development
       - ENABLE_CHROME_SANDBOX=false
     command: "celery -A attendee worker -l INFO"
+    depends_on:
```
noah-duncan: Please make it so that the minio stuff only runs if you pass an argument to docker compose. Sorry, we need to preserve backwards compatibility.

Alternatively, you could have a separate file like dev-minio.docker-compose.yaml

mateomurphy:
Hi @noah-duncan thanks for the feedback. I actually meant to open this PR on my fork and did it here by accident, it's definitely not ready to be considered for upstream merging. I'll keep testing and address the backwards compatibility concerns!

PR #932, opened 2026-07-31
feat: Add ability to disable transcription

noah-duncan:
Adds small changes on top of @solaoi 's PR here: https://github.com/attendee-labs/attendee/pull/930


solaoi (dismissed):
Thanks for following up on #930. I agree with this simplification. Treating enabled as a master switch—where false takes precedence and true or omission preserves the existing defaults—makes sense to me.

I noticed that the current test failures are from assertions that still reflect the previous validation behavior. Once those are addressed, this looks good to me. Thanks!

noah-duncan:
> Thanks for following up on #930. I agree with this simplification. Treating enabled as a master switch—where false takes precedence and true or omission preserves the existing defaults—makes sense to me.
> 
> I noticed that the current test failures are from assertions that still reflect the previous validation behavior. Once those are addressed, this looks good to me. Thanks!

Thanks for reviewing. Sorry should've made this PR a draft, it wasn't totally ready for review. 

I also wanted to get rid of the `use_one_way_audio parameter`, it's a weird name (the Zoom Meeting SDK calls per participant audio "one way audio" for some reason and I copied that) and it's not really needed in the codebase.

The awkward change you had to make to get silence detection for Zoom working even when transcription is disabled made that apparent to me.

solaoi:
Thanks for the clarification and for cleaning this up. Removing `use_one_way_audio` and deriving the behavior from the callbacks is much clearer. I reviewed the latest changes, including the Zoom silence-detection path, and everything looks good to me.

solaoi (approved):

solaoi:
@noah-duncan 
Hi! No rush at all, but since I was a bit late tagging you, I just wanted to make sure this didn’t get missed from the merge. Could you please take a look and merge it if needed?

PR #933, opened 2026-08-04
feat: add an option to pin a Google Meet message so messages can reach late joiners

littlebobert:
## Problem

Google Meet chat messages sent by a bot are only visible to participants who are present when the message is sent. Participants who join later can therefore miss important information such as meeting links, instructions, or notices.

## Solution

Add an optional `pin` field to the send-chat-message request. When `pin` is `true`, a Google Meet bot sends the message and then pins it so that late joiners can see it.

The option defaults to `false`, preserving the existing behavior for current API clients and other meeting platforms.

## example
```
{
  "to": "everyone",
  "message": "Important meeting link: https://example.com",
  "pin": true
}
```

## Validation

- `python manage.py test --tag google_meet_tests --verbosity 2` — 26 tests passed
- `ruff check .`
- `ruff format --check --diff .`

### DIFF
--- a/bots/bot_controller/bot_controller.py
+++ b/bots/bot_controller/bot_controller.py
@@ -1145,15 +1145,43 @@ class BotController:
         chat_message_requests = self.bot_in_db.chat_message_requests.filter(state=BotChatMessageRequestStates.ENQUEUED)
         for chat_message_request in chat_message_requests:
             try:
-                self.adapter.send_chat_message(
+                delivery_result = self.adapter.send_chat_message(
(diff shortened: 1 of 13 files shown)


noah-duncan:
Hi @littlebobert thanks for the PR, this is a nice feature.

I would encourage you to look into implementing this kind of thing via the network layer instead of through UI interactions. Doing it via the network layer is more reliable because it is less likely to change than the UI.

By following the technique in this PR it may be possible to pin messages via the network: https://github.com/attendee-labs/attendee/pull/523

PR #953, opened 2026-08-11
feat: Add remover fields to event_metadata

pylead:
Adds `remover_name`, `remover_uuid`, `remover_user_uuid` and `remover_is_host` to `event_metadata` on `bot.state_change`, naming the participant who denied the bot's request to join or removed it from the meeting. Microsoft Teams already names that participant in the `sender` of its conversation end message, but Attendee discarded it, so there was no way to tell a customer who stopped their bot. Only Teams reports this, so the fields are absent on Google Meet and Zoom, and absent whenever the meeting did not say who was responsible.

These are the same attributes reported for the speaker of a transcription utterance and for the participant of a participant event.

Both cases arrive as a `meeting_ended` event today, and `event_sub_type` stays null — telling them apart needs evidence we do not currently read, and is left to a follow-up branch. This PR only reports who.

Denied from the lobby, where the bot never saw the meeting's participants, so the attributes that come from a participant record are null. Teams ends the conversation as soon as the request is denied, so this arrives out of the `joining` state, not as `could_not_join_meeting` / `request_to_join_denied`. The `data` field of the webhook:

```json
{
  "new_state": "post_processing",
  "old_state": "joining",
  "created_at": "2023-07-15T14:30:45.123456Z",
  "event_type": "meeting_ended",
  "event_sub_type": null,
  "event_metadata": {
    "remover_name": "Test User",
    "remover_uuid": "8:orgid:00000000-0000-0000-0000-000000000001",
    "remover_user_uuid": null,
    "remover_is_host": null
  }
}
```

Removed after joining, where the participant is one the bot saw, so every attribute is reported:

```json
{
  "new_state": "post_processing",
  "old_state": "joined_recording",
  "created_at": "2023-07-15T14:30:45.123456Z",
  "event_type": "meeting_ended",
  "event_sub_type": null,
  "event_metadata": {
    "remover_name": "Test User",
    "remover_uuid": "8:orgid:00000000-0000-0000-0000-000000000001",
    "remover_user_uuid": null,
    "remover_is_host": true
  }
}
```

The `could_not_join_meeting` / `request_to_join_denied` event also carries the fields when the bot does reach that path, but in testing the conversation end message always got there first, so the denial surfaces as `meeting_ended`.


bots/bot_controller/bot_controller.py
```diff
@@ -1951,10 +1965,11 @@ def take_action_based_on_message_from_adapter(self, message):
         if message.get("message") == BotAdapter.Messages.MEETING_ENDED:
             logger.info("Received message that meeting ended")
             self.flush_utterances()
+            event_metadata = self.get_remover_metadata(message)
             if self.bot_in_db.state == BotStates.LEAVING:
-                new_bot_event = BotEventManager.create_event(bot=self.bot_in_db, event_type=BotEventTypes.BOT_LEFT_MEETING)
+                new_bot_event = BotEventManager.create_event(bot=self.bot_in_db, event_type=BotEventTypes.BOT_LEFT_MEETING, event_metadata=event_metadata)
```
noah-duncan: This is for the rare case where the bot was in the process of leaving when someone removed it. Do you think it is necessary to inject the metadata for this scenario? It may be confusing if the user sees "Bot left meeting - user requested" and "remover".

bots/web_bot_adapter/web_bot_adapter.py
```diff
@@ -411,6 +412,9 @@ def handle_websocket(self, websocket):
                                 self.send_message_callback({"message": self.Messages.READY_TO_SEND_CHAT_MESSAGE})
 
                         elif json_data.get("type") == "MeetingStatusChange":
+                            if json_data.get("remover") and not self.meeting_uuid_mismatch(json_data):
+                                self.remover = json_data["remover"]
```
noah-duncan: We're trying to avoid having any business logic here. Can you refactor to be
```
def handle_remover_data(json_data):
     if self.meeting_uuid_mismatch(json_data):
         return
     self.remover = json_data["remover"]
```

bots/bot_controller/bot_controller.py
```diff
@@ -1769,6 +1769,19 @@ def save_debug_artifacts_with_no_error_handling(self, message, new_bot_event):
                     save=True,
                 )
 
+    def get_remover_metadata(self, message):
+        """Event metadata naming the participant who removed the bot, when the meeting told us who it was."""
+        remover = message.get("remover")
+        if not remover:
+            return {}
```
noah-duncan: Very minor, but we should pass None here because that's the default value of create_event method

```
    @classmethod
    def create_event(
        cls,
        bot: Bot,
        event_type: int,
        event_sub_type: int = None,
        event_metadata: dict = None,
        max_retries: int = 3,
    ) -> BotEvent:
```

docs/webhooks.md
```diff
@@ -102,9 +102,40 @@ For webhooks triggered by `bot.state_change`, the `data` field contains:
   "created_at": < The timestamp when the state change occurred >,
   "event_type": < The type of event that triggered the state change >,
   "event_sub_type": < The sub-type of event that triggered the state change >,
+  "event_metadata": < Extra details about the event, when there are any >,
 }
 ```
 
+### Knowing who removed the bot
```
noah-duncan, line 125: Can you put this below the `Using webhooks to know when the recording is available` section, since this feature is more niche?

docs/webhooks.md
```diff
@@ -102,9 +102,40 @@ For webhooks triggered by `bot.state_change`, the `data` field contains:
...
+  "event_metadata": {
+    "remover_name": "Test User",
+    "remover_uuid": "8:orgid:00000000-0000-0000-0000-000000000001",
+    "remover_user_uuid": null,
+    "remover_is_host": true
+  }
+}
+```
+
+These are the same attributes reported for the speaker of a transcription utterance, for the participant of a participant event, and by `GET /api/v1/bots/{object_id}/participants`.
```
noah-duncan: Don't think you need this sentence

docs/webhooks.md
```diff
@@ -102,9 +102,40 @@ For webhooks triggered by `bot.state_change`, the `data` field contains:
   "created_at": < The timestamp when the state change occurred >,
   "event_type": < The type of event that triggered the state change >,
   "event_sub_type": < The sub-type of event that triggered the state change >,
+  "event_metadata": < Extra details about the event, when there are any >,
 }
 ```
 
+### Knowing who removed the bot
+
+When a participant denies the bot's request to join, or removes the bot from the meeting,
```
noah-duncan, line 129: State up-front that this is only supported for teams

docs/webhooks.md
```diff
@@ -102,9 +102,40 @@ For webhooks triggered by `bot.state_change`, the `data` field contains:
...
+  }
+}
+```
+
+These are the same attributes reported for the speaker of a transcription utterance, for the participant of a participant event, and by `GET /api/v1/bots/{object_id}/participants`.
+
+A few things to know about these fields:
+
+- Only Microsoft Teams reports who removed the bot. Google Meet and Zoom tell the bot that it was removed but not by whom.
+- `remover_user_uuid` and `remover_is_host` are null when that participant was not in the meeting with the bot. When the bot is denied from the lobby it never sees the meeting's participants, so only `remover_name` and `remover_uuid` are known.
```
noah-duncan: Dunno if this is worth mentioning

docs/webhooks.md
```diff
@@ -102,9 +102,40 @@ For webhooks triggered by `bot.state_change`, the `data` field contains:
...
+- The fields are absent whenever the bot was not told who was responsible. Their absence is not evidence that nobody removed the bot.
```
noah-duncan: Dunno if this is worth mentioning

bots/bot_controller/bot_controller.py
pylead: Dropped it for that branch. A bot already in the leaving state decided to leave on its own — /leave, or auto-leave on silence — so naming a remover next to "user requested" would only confuse the reader. The metadata is now only injected on the meeting_ended branch (and on request_to_join_denied), with a test guarding the leaving case. 3bffd021

bots/bot_controller/bot_controller.py
pylead: Returns None now. 06b2b1bc

bots/web_bot_adapter/web_bot_adapter.py
pylead: Moved into handle_remover_data, next to the other handlers, and the loop just calls it. One difference from your snippet: I kept the remover check inside the handler rather than calling meeting_uuid_mismatch unconditionally, because that method sets self.meeting_uuid when it is unset, so calling it for every MeetingStatusChange would change when the meeting uuid gets pinned. Assigning unconditionally would also clear an already-captured remover when a later message carries none, and send_request_to_join_denied_message reads self.remover some time after it is set.

```python
def handle_remover_data(self, json_data):
    # A meeting status change names the participant who removed us when it knows who that was.
    if not json_data.get("remover") or self.meeting_uuid_mismatch(json_data):
        return

    self.remover = json_data["remover"]
```

883042f0

docs/webhooks.md
pylead: Moved below it. b303839d

docs/webhooks.md
pylead: The section now opens with it: "This is only supported for Microsoft Teams. Google Meet and Zoom tell the bot that it was removed but not by whom."

docs/webhooks.md
pylead: Removed.

docs/webhooks.md
pylead: Removed.

docs/webhooks.md
pylead: Removed.

noah-duncan (changes requested):
Small thing, but the `delete_data` method leaves the webhook delivery attempts for the bot.state_change trigger because it is assumed there is no PII associated with these. See the code here: https://github.com/attendee-labs/attendee/blob/79692902618bc2b989475ff1b59bfa4d06b9aa78/bots/models.py#L904

Due to the addition of remover_x attributes, this is no longer true.

So I would scrub the remover attributes from the metadata, on both the bot_event itself and webhook delivery attempt.

noah-duncan (approved):
Thanks @pylead !

PR #957, opened 2026-08-13
Support configuring database via separate POSTGRES_* env vars

ppeerttu:
## Summary

Allows configuring the production/staging database connection with individual environment variables as an alternative to a full `DATABASE_URL`:

- `POSTGRES_HOST` (required for this mode)
- `POSTGRES_DB` (required for this mode)
- `POSTGRES_USER` (required for this mode)
- `POSTGRES_PASSWORD` (optional)
- `POSTGRES_PORT` (optional, defaults to `5432`)

This is useful when credentials are injected as separate secrets (e.g. Kubernetes secrets, Cloud SQL, various secret managers) and composing a full connection URL is awkward.

## Backwards compatibility

`DATABASE_URL` still takes precedence whenever it is set, so existing deployments are unaffected. The individual variables are only consulted when `DATABASE_URL` is unset.

## Changes

- New `attendee/settings/db.py` with a shared `default_database(ssl_require)` helper that builds the connection from `DATABASE_URL` or the `POSTGRES_*` vars. Credentials and database name are URL-escaped, so passwords with special characters work. If neither form is configured, it raises `ImproperlyConfigured` with a clear message instead of failing later with an empty database config.
- `production.py`, `production-gke.py` and `staging-gke.py` now use the shared helper instead of three duplicated `dj_database_url.config(...)` blocks. Per-environment behavior is unchanged (`POSTGRES_SSL_REQUIRE`-driven SSL in `production.py`, always-required SSL in the GKE settings, `DISABLE_SERVER_SIDE_CURSORS`, `conn_max_age=600`, health checks).
- Documented the new variables in `docs/environment-variables.md`.

## Testing

Verified the helper in isolation: the `POSTGRES_*` fallback produces the expected Django config (including `sslmode=require` and special characters in passwords round-tripping correctly), `DATABASE_URL` wins when both are present, and a missing configuration raises the explicit error.

### DIFF
--- a/attendee/settings/db.py
+++ b/attendee/settings/db.py
@@ -1,0 +1,36 @@
+import os
+from urllib.parse import quote
+
+import dj_database_url
+from django.core.exceptions import ImproperlyConfigured
+
+POSTGRES_ENV_VARS = ("POSTGRES_HOST", "POSTGRES_DB", "POSTGRES_USER")
+
+
+def _database_url_from_postgres_env():
+    # DATABASE_URL takes precedence; these vars are only consulted when it's unset.
+    host = os.getenv("POSTGRES_HOST")
+    name = os.getenv("POSTGRES_DB")
+    user = os.getenv("POSTGRES_USER")
+    if not (host and name and user):
+        return None
+    password = os.getenv("POSTGRES_PASSWORD", "")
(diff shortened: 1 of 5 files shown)


noah-duncan:
Closing in order to make a PR with a version bump included

PR #961, opened 2026-08-17
Delete debug files when deleting bot data

### LINKED ISSUE #960: delete_data leaves debug screen recordings (meeting content) in object storage
Observed on a self-hosted deployment (dev compose stack) at current `main` (`94d958d`).

## Summary

`POST /api/v1/bots/{id}/delete_data` reports success and the bot reaches `data_deleted`, but `debug_screen_recording_*.mp4` and `debug_screenshot_*` objects remain in the recordings bucket. Debug screen recordings capture the bot's view of the meeting, so meeting content outlives data deletion. For self-hosters who rely on `delete_data` for retention/GDPR erasure, this is a silent gap.

## Cause

Two parts:

1. `Bot.delete_data()` removes debug screenshots with a queryset bulk delete (`bots/models.py:881`):

   ```python
   BotDebugScreenshot.objects.filter(bot_event__bot=self).delete()
   ```

   `BotDebugScreenshot.file` is a `FileField` (`bots/models.py:3035`), and a queryset `.delete()` never calls the storage backend, so the objects stay in the bucket. Contrast with the recording path a few lines below, which calls `recording.file.delete()` explicitly.

2. The impact is amplified because `Bot.create_debug_recording()` force-enables debug screen recording for every Google Meet / Teams / Zoom-web bot with audio+video recording, regardless of `SAVE_DEBUG_RECORDINGS` and `debug_settings` (`bots/models.py:1159`):

   ```python
   # Temporarily enabling this for all google meet meetings
   bot_meeting_type = meeting_type_from_url(self.meeting_url)
   if (bot_meeting_type == MeetingTypes.GOOGLE_MEET or ...) and self.recording_type() == RecordingTypes.AUDIO_AND_VIDEO:
       return True
   ```

   So essentially every Meet bot produces a debug screen recording that `delete_data` then leaves behind.

## Reproduction

1. Self-hosted stack with S3-compatible storage (MinIO in my case).
2. Create a bot for a Google Meet meeting with default settings; let it join and record; end the meeting.
3. Wait for `ended`, then `POST /api/v1/bots/{id}/delete_data` → 200, state `data_deleted`, recording file is gone from the bucket.
4. List the bucket: `debug_screen_recording_<id>.mp4` (and any `debug_screenshot_<id>.*`) objects are still there, now orphaned — their DB rows are gone, so nothing references or ever deletes them.

## Suggested fix

In `delete_data()`, iterate the debug screenshots and delete their files before removing the rows, mirroring the recording handling:

```python
for screenshot in BotDebugScreenshot.objects.filter(bot_event__bot=self):
    if screenshot.file and screenshot.file.name:
        screenshot.file.delete()
BotDebugScreenshot.objects.filter(bot_event__bot=self).delete()
```

Separately, it would be good if the forced-debug-recording branch respected `SAVE_DEBUG_RECORDINGS`/`debug_settings`, or was at least documented — as written, a self-hoster who sets `SAVE_DEBUG_RECORDINGS=false` still records meeting content to storage on every Meet bot.

Happy to send a PR for the `delete_data()` part if useful.

Nikola-Milovic:
## Summary

- delete every debug screenshot, MHTML capture, screen recording, and bot-log object from configured storage before removing its database row
- serialize debug-artifact uploads with bot data deletion so a cleanup upload cannot recreate sensitive data after deletion succeeds
- make deletion retry-safe after partial storage failures and idempotent once the bot reaches `DATA_DELETED`
- route all six production debug-artifact write paths through the guarded creation helper
- add regression and real PostgreSQL concurrency coverage for both upload-before-delete and delete-before-upload orderings

## Root cause

`QuerySet.delete()` removed `BotDebugScreenshot` rows without invoking the `FileField` storage backend, which left debug artifacts orphaned in object storage. The initial fix also needed protection against cleanup writers racing with the deletion sweep and uploading a new object after deletion.

## Impact

`delete_data()` removes the selected bot's debug artifacts from object storage while preserving other bots' objects. Late artifact creation is rejected, a directly uploaded late object is cleaned up, and repeated deletion converges safely without creating duplicate `DATA_DELETED` events.

The bot row lock intentionally spans storage I/O to guarantee that deletion cannot report success while a debug upload for the same bot is still active. This trades longer per-bot lock duration for deterministic deletion semantics.

## Validation

- 18 targeted Django tests passed against local PostgreSQL and Redis, including 13 bot-data-deletion tests, four hard-deletion tests, and the bot-log attachment integration test
- real MinIO integration passed: target recording and debug objects deleted, control-bot objects preserved, late upload removed, post-delete creation blocked, empty artifact row removed, and repeated deletion succeeded
- `ruff check` passed for all three changed files
- `ruff format --check` passed for all three changed files
- `python -m py_compile` passed for all three changed files
- `git diff --check` passed
- Claude Code CLI 2.1.234 (Sonnet) final review: no blocking findings; only the documented lock-across-storage-I/O tradeoff

Closes #960

### DIFF
--- a/bots/bot_controller/bot_controller.py
+++ b/bots/bot_controller/bot_controller.py
@@ -1681,11 +1681,13 @@ class BotController:
         # Find the bot's last event
         last_bot_event = self.bot_in_db.last_bot_event()
         if last_bot_event:
-            debug_screenshot = BotDebugScreenshot.objects.create(bot_event=last_bot_event)
-
             # Save the file directly from the file path
             with open(BotAdapter.DEBUG_RECORDING_FILE_PATH, "rb") as f:
-                debug_screenshot.file.save(f"debug_screen_recording_{debug_screenshot.object_id}.mp4", f, save=True)
+                debug_screenshot = BotDebugScreenshot.create_with_file(
+                    bot_event=last_bot_event,
+                    filename_template="debug_screen_recording_{object_id}.mp4",
+                    content=f,
+                )
             logger.info(f"Saved debug recording with ID {debug_screenshot.object_id}")
 
     def save_bot_pod_logs(self):
@@ -1718,9 +1720,12 @@ class BotController:
             return
 
         for part_number, path in enumerate(log_file_paths_in_order, start=1):
-            debug_screenshot = BotDebugScreenshot.objects.create(bot_event=last_bot_event)
             with open(path, "rb") as f:
-                debug_screenshot.file.save(f"bot_logs_part_{part_number}_{debug_screenshot.object_id}.log", f, save=True)
+                debug_screenshot = BotDebugScreenshot.create_with_file(
+                    bot_event=last_bot_event,
+                    filename_template=f"bot_logs_part_{part_number}_{{object_id}}.log",
+                    content=f,
+                )
             logger.info(f"Saved bot pod logs from {path} with ID {debug_screenshot.object_id}")
 
     def on_message_from_websocket_audio(self, message_json: str):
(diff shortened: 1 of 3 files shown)


noah-duncan:
@Nikola-Milovic let me know your thoughts on this alternate solution for the races:
https://github.com/attendee-labs/attendee/pull/966

Also we will add the ability configure a separate s3 bucket for debug screenshots. So as a different safeguard you can simply set that bucket to only retain files for a very short time.

Nikola-Milovic:
Looks good, I'll close my PR

PR #984, opened 2026-08-28
fix: don't let a hung chromedriver stall joins or bot shutdown

rafaeljusto:
`init_driver()`, `abort_join_attempt()` and `cleanup()` all call `close()` then `quit()` on the driver. Against an unresponsive chromedriver each call blocks for selenium's 120s timeout times `urllib3`'s three retries, so a single teardown can burn ~12 minutes. The cost differs per call site:

- `init_driver()` — the retry can't start a new browser, so the bot misses most of the meeting it was retrying to join.
- `abort_join_attempt()` — the "could not join" report is delayed by minutes.
- `cleanup()` — a finished bot holds its pod open long after the meeting ended.

All three now run on a thread that is abandoned after 30s (`run_driver_call_with_timeout`), after which the chromedriver process is killed so nothing is left hanging. Healthy drivers are unaffected: the kill only fires when the graceful teardown is still blocked. `cleanup()` also runs `log_browser_history()` and the subclass pre-close hook inside the bounded region, since those touch the driver too.

Covered by `bots/tests/test_driver_teardown.py`.

Not covered here: `cleanup()` makes one driver call (`disableMediaSending`) before this bounded section, and the adapters make many others during a join. Bounding every driver interaction is a larger change — this PR covers the teardown paths where a hang is both likely and costly.


noah-duncan (approved):
Thanks @rafaeljusto !

PR #986, opened 2026-08-31
fix: optionally log Celery queue sizes

pylead:
The scheduler used to log the size of every Celery queue on each cycle. Those logs were dropped in
- #970

when the instance health snapshots landed, but they're still useful when watching a backlog live.

This brings them back behind `LOG_CELERY_QUEUE_SIZES`, which defaults to `false`, so nothing is logged unless it's turned on.

The sizes come from `get_celery_queue_depths()`, the helper the health snapshot taker already uses, rather than a second copy of the queue-name derivation.

Tests cover the flag on, the flag off, and a failed read.


noah-duncan:
Not needed because this logs the same info https://github.com/attendee-labs/attendee/pull/987

PR #1006, opened 2026-09-10
fix: allow reading incomplete async transcription

pylead:
### Story

Allow reading incomplete by status `utterances_still_in_progress_when_transcription_terminated` transcriptions.
In fact they are frequently complete after some time, but remain `failed` on attendee.

### Tecnical details

Currently the request-response pair looks like this:

`GET /bots/{bot_id}/transcript?async_transcription_id={trans_id}`
`400 {"error": "... is not complete. It is in state failed"}`.

The text is unreachable, and the only way out is a second transcription of the same recording.

The fix allows to get that incomplete transcription if it's available:

`GET /bots/{bot_id}/transcript?async_transcription_id={trans_id}&include_incomplete=true`
`400`
```json
{
  "error": "Async transcription tran_xxx is not complete. It is in state failed",
  "incomplete_transcription": [
    {"speaker_name": "...", "timestamp_ms": 1788965971473, "duration_ms": 3100,
     "transcription": {"transcript": "..."}}
  ]
}
```

Without the parameter nothing changes. It applies to `in_progress` and `not_started` too, since a caller that asks for an incomplete transcript should get whatever exists. `updated_after` and `split_on_turns` apply to the returned list as usual.


pylead:
We decided to always get the transcription - doesn't matter complete or incomplete and return it as success response.
Instead, we rely on webhooks to know if it's complete.

PR #1007, opened 2026-09-10
feat: make scheduler past join_at tolerance configurable

ppeerttu:
The scheduler skips scheduled bots whose `join_at` is more than 5 minutes in the past. That window is hardcoded, so a scheduler outage longer than 5 minutes causes bots to be missed.

Add `SCHEDULED_BOT_PAST_JOIN_AT_TOLERANCE_SECONDS` to make the window configurable. Default stays at 300 seconds. Applies to both the plain and jitter code paths.


ppeerttu:
Superseded by https://github.com/attendee-labs/attendee/pull/1009, which uses a branch in this repo instead of a fork.

PR #1008, opened 2026-09-10
fix: only kill worker process group on shutdown when bots run in the celery worker

ppeerttu:
## Problem

`bots/tasks/run_bot_task.py` connects a `worker_shutting_down` handler that SIGTERMs the worker's whole process group. Celery's response to SIGTERM is a warm shutdown (stop consuming, finish in-flight tasks), but this handler fires at the start of it and kills every prefork child immediately. Busy children die with `WorkerLostError`, and with `acks_late` their tasks are requeued and re-run on another worker. On Kubernetes the pod's `terminationGracePeriodSeconds` is never used.

For `process_utterance_group_for_async_transcription` this means the AssemblyAI job is submitted twice and the original transcript is left orphaned at AssemblyAI. We see this several times a day on node consolidation.

The handler only makes sense when bots run inside the worker (`LAUNCH_BOT_METHOD` unset or `celery`), where a bot task spawns Chrome and other subprocesses that must not outlive the worker. With `kubernetes` and `docker-compose-multi-host`, `run_bot` never executes in the worker.

## Change

- Add `BOTS_RUN_IN_CELERY_WORKER` setting, computed from the same `LAUNCH_BOT_METHOD` check already used for `CELERY_WORKER_MAX_TASKS_PER_CHILD`.
- Only connect the `worker_shutting_down` handler when that setting is true. Behaviour is unchanged for in-process bot mode; `kill_child_processes` is untouched.
- No changes to `acks_late`, prefetch or `reject_on_worker_lost`.

## Verification

- New tests reload the task module under `override_settings` and assert the receiver is connected only when `BOTS_RUN_IN_CELERY_WORKER` is true.
- Ran in the worker image with `LAUNCH_BOT_METHOD=kubernetes` and `celery`: receiver absent / present respectively, `CELERY_WORKER_MAX_TASKS_PER_CHILD` unchanged.
- No other code references `worker_shutting_down`, `killpg` or `kill_child_processes`.

### DIFF
--- a/attendee/settings/base.py
+++ b/attendee/settings/base.py
@@ -203,7 +203,12 @@ CELERY_TASK_ROUTES = {
     },
 }
 
-if os.getenv("LAUNCH_BOT_METHOD") != "kubernetes" and os.getenv("LAUNCH_BOT_METHOD") != "docker-compose-multi-host":
+# With the kubernetes and docker-compose-multi-host launch methods, bots run in a separate pod/container
+# via `python manage.py run_bot` and the Celery worker only runs short scheduling, webhook and transcription
+# tasks. Otherwise (unset or "celery") the run_bot task itself executes inside the worker.
+BOTS_RUN_IN_CELERY_WORKER = os.getenv("LAUNCH_BOT_METHOD") not in ("kubernetes", "docker-compose-multi-host")
+
+if BOTS_RUN_IN_CELERY_WORKER:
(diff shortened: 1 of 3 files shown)


ppeerttu:
Superseded by https://github.com/attendee-labs/attendee/pull/1010 (same branch, now pushed to this repo instead of a fork).

PR #1019, opened 2026-09-14
fix: return incomplete transcript of a failed async transcription

pylead:
## Story

Ref [slack thread :thread: ](https://mailmaestro.slack.com/archives/C09EVTW6F0X/p1789065038118239?thread_ts=1789027174.095049&cid=C09EVTW6F0X)

### Why

When an async transcription ends in the `failed` state, most of the transcript is usually still there. The common case is `utterances_still_in_progress_when_transcription_terminated`: the transcription times out while a handful of utterances are still with the provider, so the whole transcription is marked failed even though nearly all of the text came back fine.

Today the transcript endpoint refuses to return anything for such a transcription, so from an API client's point of view the transcript simply doesn't exist — even though it does, and is good enough to use.

### What changes

`GET /api/v1/bots/<id>/transcript?async_transcription_id=...` now returns the transcript for a `failed` async transcription, not just a `complete` one. Clients get back whatever was transcribed and can decide for themselves whether it's long enough to keep or worth re-transcribing.

Transcriptions that haven't finished yet (`not_started`, `in_progress`) still return an error — there's nothing useful to hand back while the provider is still working, and clients should wait for the async transcription webhook.

Since a partial transcript now comes back with a 200, the transcription's actual state is learned from the async transcription state-change webhook rather than from this endpoint.


noah-duncan (approved):
Thanks @pylead

